PR #1, opened 2017-08-31
Update images.yml

bachradsusi:
Fedora 24 is EOL since 2017-08-08
Fedora 26 was released on 2017-07-11


larskarlitski:
This is from before we changed over to the new system. I've incorporated the new URLs on the testers in OpenShift. Thanks!

PR #2, opened 2017-09-04
Install python2-avocado-plugins-varianter-yaml-to-mux in container

bachradsusi:
Plugins were split into subpackages in python2-avocado-46.0-2

Fixes: avocado run: error: unrecognized arguments: -m image:images.yml


larskarlitski:
This is from before we switched to the new way of testing. We don't use avocado anymore. Sorry and thanks!

PR #3, opened 2017-09-08
Turn testing inside out

larskarlitski:
Now that we have standard-test-roles in Fedora, we can use that to
create virtual machines for us.

The role repositories don't need to include `run-tests.sh` anymore, and
as such don't depend on a random container. Instead, integration tests
can be run like this:

    export TEST_SUBJECTS=/path/to/image.qcow2
    ansible-playbook -i /usr/share/ansible/inventory/standard-inventory-qcow2 test/test_*.yml

We still need some way to run the tests on open pull requests. We used
semaphore for this before, but that has shortcomings. For one, each
repository needs to copy the test script and cannot share already
downloaded images with other repositories. Also, we cannot test against
private images (such as RHEL) on semaphore.

Thus, this commit changes most of the contents of this repository to
turn it into a test runner that can be run on a private OpenShift
instance. It polls GitHub for open pull requests, runs the tests as
above for multiple images, copies the results to an ssh server, and sets
GitHub's commit status.


README.md
```diff
@@ -1,42 +1,74 @@
...
-roles](https://github.com/linux-system-roles) against various operating
-systems.
-
-It starts a virtual machine based on a cloud image of each operating system
-(defined in [images.yml](/images.yml)) and runs playbooks in the roles `test`
-directory against that machine.
+A docker container which runs integration tests for open pull requests on
+[linux system roles](https://linux-system-roles.github.io) repositories. It runs
+all playbooks `test/test_*.yml` of the repository against various virtual
+machines.
```
stefwalter: The intent of the standard test invocation spec is to execute ```tests.yml```. Just checking in case that wasn't clear.

README.md
```diff
@@ -1,42 +1,74 @@
...
 
 To build the container, run
 
-    docker build --tag cockpit/linux-system-roles-test .
+    docker build --tag linux-system-roles-test .
+
+Multiple of these containers can be started in parallel. They try to not step
+on each others feet by synchronizing via GitHub commit statuses.
+
+The test runner needs a `config.json` in the `/config` mount, which specifies
```
stefwalter: Is there such a well-known ```config.json```? Worth linking here if so.

standard-inventory-qcow2
```diff
@@ -0,0 +1,263 @@
...
+import shlex
+import signal
+import socket
+import subprocess
+import sys
+import tempfile
+import time
+import traceback
+import distutils.util
+
```
stefwalter: It's worth including a comment here that this is a copy from standard-test-roles with patches, and eventually shouldn't be included in this repo.

test/run-tests
```diff
@@ -0,0 +1,366 @@
...
+        sys.stdout = self.f
+        sys.stderr = self.f
+
+        return self.f
+
+    def __exit__(self, *exception):
+        self.f.close()
+
+        sys.stdout = self.oldout
+        sys.stderr = self.olderr
```
stefwalter: The close should happen after setting stdout/stderr back. In addition there should be flush() calls here.

README.md
larskarlitski: Yeah, I know. I wanted to have something that works with the repositories that we have now.

But now that you're saying it, it probably makes sense to support both right away. I've changed it to look for `tests/tests.yml` and fall back to the old way only if that doesn't exist.

README.md
larskarlitski: There is, right above in the `README.md`. Do you think it makes sense to add that as an `example-config.json` in the root of the repository?

I didn't include the one I'm using on OpenShift right now because that includes links to internal RHEL images.

standard-inventory-qcow2
larskarlitski: Done.

test/run-tests
larskarlitski: Closing the file implies flushing, doesn't it? (Note that I am flushing `stdout` and `stderr` before redirecting, which was the main motivation of making this instead of using `contextlib.redirect_stdout`).

 Why is closing before setting them back preferable?

bachradsusi:
The scenario described in the commit message works for me. There's only a problem with `.qcow2c` images -  `standard-inventory-qcow2` handles only `.qcow2` subjects

larskarlitski:
> The scenario described in the commit message works for me. There's only a problem with .qcow2c images - standard-inventory-qcow2 handles only .qcow2 subjects

Yeah I noticed. `qcow2c` is a weird extension though. I've only ever seen CentOS images use it. We can add it to standard-test-roles or just use the `.qcow` images for CentOS (which I'm doing on the instance I have running right now).

larskarlitski:
> I would be for adding qcow2c support to standard-inventory-qcow ... if we indeed have such images to run tests on. Do you agree @mmathesius

I've opened [pull request #56](https://pagure.io/standard-test-roles/pull-request/56) for standard-test-roles so that we can have the discussion over there.

larskarlitski:
> The intent of the standard test invocation spec is to execute tests.yml. Just checking in case that wasn't clear.

@stefwalter: I've changed it (see comment above) to use `tests/tests.yml`, but that means that all tests get executed against the same instance of the virtual machine. Is that what we want? I think I'd rather be able to have completely independent tests.

mmathesius:
>> I would be for adding qcow2c support to standard-inventory-qcow ... if we indeed have such images to run tests on. Do you agree @mmathesius
> 
> I've opened pull request \#56 for standard-test-roles so that we can have the discussion over there.

Yes. The [standard-test-roles pull request #56](https://pagure.io/standard-test-roles/pull-request/56) has been verified and merged.

PR #4, opened 2017-10-02
run-tests: run tests/tests*.yml

larskarlitski:
This is a recent (backwards-compatible) change in the fedora testing
spec:

  https://fedoraproject.org/wiki/Changes/InvokingTests#Invocation


test/run-tests
```diff
@@ -180,16 +180,13 @@ class Task:
                 with open(setup_file, 'w') as f:
                     json.dump([play], f)
 
-                playbooks.append(setup_file)
-
             # Fedora's standard test invocation spec mandates a
```
tyll: this needs to be changed to

    mandates running all playbooks matching

to match to the change from running just one playbook to several.

larskarlitski: Fixed. Thanks.

tyll:
Thank you. LGTM. What is the process here? Should I go ahead and merge or is there someone else that needs to approve this PR, too?

PR #7, opened 2018-05-14
Report failure when no test playbooks are found

tyll:
This avoids false-positives if for some reason the playbooks are renamed
to invalid names. Otherwise missing playbooks are reported as success
even when no tests ran.

### DIFF
--- a/test/run-tests
+++ b/test/run-tests
@@ -185,9 +185,15 @@ class Task:
             # used to be tested by running all playbooks `test/test_*.yml`.
             # Support both, but prefer the standard way. Can be removed once
             # all repos are moved over.
-            playbooks = glob.glob(f'{sourcedir}/tests/tests*.yml')
+            playbookglob = f'{sourcedir}/tests/tests*.yml'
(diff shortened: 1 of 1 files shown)


test/run-tests
```diff
@@ -185,10 +185,16 @@ class Task:
...
             # all repos are moved over.
-            playbooks = glob.glob(f'{sourcedir}/tests/tests*.yml')
+            playbookglob = f'{sourcedir}/tests/tests*.yml'
+            playbooks = glob.glob(playbookglob)
             if not playbooks:
                 playbooks = glob.glob(f'{sourcedir}/test/test_*.yml')
 
+            if not playbooks:
+                print(f'No test playbooks found, please add at least one'
+                      'playbook that matches {playbookglob}.')
```
pcahyna: The problem is that some roles have no tests.

pcahyna: Could be solved by excluding those (which don't have tests yet) from the list of repositories in `config.json`.

PR #19, opened 2018-05-23
Add command line argument support to make it easier to run locally

tyll:
This will also make it easier to run against single PRs.


tyll:
@larskarlitski could you maybe review the new changes, too? It would be great to get this merged today or tomorrow, afterwards I plan to focus on another task.

README.md
```diff
@@ -20,17 +20,32 @@ accessible URL for these results. For example:
...
+    "linux-system-roles/kdump",
+    "linux-system-roles/firewall",
+    "linux-system-roles/postfix",
   ],
   "images": [
     {
-      "name": "fedora-26",
-      "source": "https://download.fedoraproject.org/pub/fedora/linux/releases/26/CloudImages/x86_64/images/Fedora-Cloud-Base-26-1.5.x86_64.qcow2",
+      "name": "fedora-27",
+      "source": "https://download.fedoraproject.org/fedora/linux/releases/27/CloudImages/x86_64/images/Fedora-Atomic-27-1.6.x86_64.qcow2",
```
larskarlitski: Is it intentional that you used the atomic image here? I think it might make sense to test on atomic, but this was supposed to be the standard fedora image.

pcahyna: This was my question as well. Also, the image does not seem to be accessible at that URL.

tyll: Awesome, you found a bug. I will fix it.

test/run-tests
```diff
@@ -377,8 +382,16 @@ def main():
...
+        "--inventory", help="Inventory to use for VMs",
         default='/usr/share/ansible/inventory/standard-inventory-qcow2')
+    parser.add_argument("pull_request", nargs="*", default=None,
+                        help="Pull requests to test. Example: "
+                        "linux-system-roles/network/1")
+    parser.add_argument(
+        "--use-images", default="*",
+        help="Test pull request only against images matching this pattern")
+    parser.add_argument("--dry-run", help="Do not update pull request status",
+                        default=False, action="store_true")
```
pcahyna: It seems that with `--dry-run` the code still needs `/secrets` although it maybe should not?

test/run-tests
```diff
@@ -377,8 +382,16 @@ def main():
     parser.add_argument("--cache", default="/cache",
                         help="Directory for caching VM images")
     parser.add_argument(
-        "--inventory",
+        "--inventory", help="Inventory to use for VMs",
         default='/usr/share/ansible/inventory/standard-inventory-qcow2')
+    parser.add_argument("pull_request", nargs="*", default=None,
+                        help="Pull requests to test. Example: "
```
pcahyna: Is it intentional that it is `pull_request` and not `--pull_request`?

pcahyna: Ah, it is intended to run like `run-tests linux-system-roles/network/50`, right?

test/run-tests
pcahyna: Also `--dry-run` apparently also means that the test artifacts are not uploaded.

tyll: `--dry-run` still needs to use the Github API to get the details of the pull request, therefore I guess it needs a Github token from the secrets. Not sure if it would be possible to get the necessary details without authentication to the API, it might be possible.

Are you suggesting that the artifacts should be uploaded or that it should be mentioned in the help?

test/run-tests
tyll: yes, this is the intended command line

test/run-tests
pcahyna: > --dry-run still needs to use the Github API to get the details of the pull request, therefore I guess it needs a Github token from the secrets. Not sure if it would be possible to get the necessary details without authentication to the API, it might be possible.

It would be great if it were possible, it would make local running of the test harness more easy and useful. Not sure whether it is possible to fetch this information anonymously though.

> Are you suggesting that the artifacts should be uploaded or that it should be mentioned in the help?

The latter.

pcahyna: https://developer.github.com/v3/#authentication says "Requests that require authentication" which seems to imply that there are requests which don't. I have not tried it though.

test/run-tests
```diff
@@ -339,49 +366,160 @@ def main():
...
+            print(traceback.format_exc())
+            state = 'error'
+            description = 'Exception when running tests: ' + str(e)
+
+    run('chmod', 'a+rX', workdir)
+    results_dir = f'{task.owner}-{task.repo}-{task.id_}-{timestamp}'
+
+    if dry_run:
+        print(f"Artifacts kept at: {artifactsdir}")
+        with open("{artifactsdir}/test.log") as test_log:
```
pcahyna: format is missing here.

pcahyna:
Great, it is working now!

PR #26, opened 2018-06-25
Allow to specify arbitrary Ansible playbook in config.json.

pcahyna:
This generalizes the images[i].setup command in config.json.

The playbook shall be given as a JSON array under the images[i].setup_plays key in config.json. This is possible because Ansible playbooks are YAML and can be represented as JSON.

Eventually, images[i].setup can be removed.


tyll:
Can you maybe add an example to the README? How about some checks for images[i].setup to check if it is just a string and do what happens now and if it is a list of add them as plays? Then we can still use setup  and it should not be much more code.

pcahyna:
The new commit addresses both of your comments.

Issue #23, opened 2018-05-30
allow to whitelist commits instead of PRs for CI runs


tyll:
Currently only PRs can be whitelisted with the `needs-ci` label but I suppose this would also whitelist new pushes/commits to the PR after the label was added. Maybe it would be better to use a comment and only check commits submitted before the commit with the CI system. Not sure if there is another possiblity.

PR #27, opened 2018-06-29
Allow to whitelist commits in PR to run CI

tyll:
- Use a comment like ci-check-commit:<commit hash> to whitelist a commit
- This addresses #23


larskarlitski:
Reworded your commit message a bit to include "Fixes #23", so that github picks it up. Thanks!

PR #29, opened 2018-07-06
Run dnf commands together and remove cache

tyll:
The way docker caches images, it is recommended to remove the package
cache to ensure that later steps do not use outdated cache files and to
reduce the image size. To avoid re-downloads of the cache, run all
commands at once.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,10 @@
 FROM fedora:28
 
-RUN dnf update -y
-
-RUN dnf install -y \
+RUN dnf update -y && dnf install -y \
     ansible \
     git \
     python3-requests \
-    standard-test-roles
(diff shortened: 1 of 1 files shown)


pcahyna:
Does it change anything to merge the two RUN statements into one?

tyll:
From what I learned and saw about docker, each RUN statement will create a cached layer, therefore it is better to remove the cached metadata after running DNF, otherwise old metadata would be cached. Merging both dnf commands removes the need to download the metadata again for both layers, when the metada would be removed for both commands.

I do not know yet, whether images built by the docker hub will be done with or without caching but I noticed that for running it locally, I need to explicitly disable caching for the layers to be actually updated.

PR #30, opened 2018-07-07
Allow to specify pull requests with URLs

tyll:


test/run-tests
```diff
@@ -512,11 +512,21 @@ def main():
...
+            github_baseurl = 'https://github.com/'
+            # remove protocol and hostname
+            pull_request = pull_request.split(github_baseurl, 1)[-1]
+            # Specifications as copied from a pull request URL:
+            # Example: linux-system-roles/network/pull/67
+            owner, repo, _, pullnr = pull_request.split('/')
+        else:
+            # Short version:
+            # Example: linux-system-roles/network/67
+            owner, repo, pullnr = pull_request.split('/')
```
larskarlitski: Please use `urllib.parse.urlparse` to parse the URL. This will work for the "short version" as well. Something like this:
```python
o = urllib.parse.urlparse(args.pull_request)
owner, repo, pullnr = o.path.strip('/').replace('/pull/', '/').split('/')
```

tyll: Good idea, I changed it.

PR #33, opened 2018-07-10
Use code style: black and add tox/travis config

tyll:


tox.ini
```diff
@@ -0,0 +1,31 @@
...
+commands =
+    flake8 --statistics {posargs} \
+    {[base]scriptfiles}
+
+[flake8]
+show_source = True
+max-line-length = 88
+
+[travis]
+python = 3.6: black,flake8
```
pcahyna: ERROR:   py37: Error creating virtualenv. ... Error details: InvocationError('Failed to get version_info for python3.6: b"pyenv: python3.6: command not found"
Could it be related to 3.6 here vs.  3.7-dev in the first hunk?

tyll: I removed 3.7-dev from travis for now since we do not have any unit tests that would make sense to run against different python versions.

PR #40, opened 2018-08-20
Abort if there is not enough disk space

tyll:
This should kill PODs automatically if they cannot work because of missing disk space and make OpenShift spin up another one.

### DIFF
--- a/test/run-tests
+++ b/test/run-tests
@@ -185,7 +185,7 @@ class Task:
         self.image = image
         self.inventory = "/usr/share/ansible/inventory/standard-inventory-qcow2"
 
-        self.id_ = f"pull-{self.pull}-{self.head[:7]}-" + self.image["name"]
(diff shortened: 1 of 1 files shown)


test/run-tests
```diff
@@ -578,6 +586,10 @@ def handle_task(gh, args, config, task, dry_run=False):
         )
     print()
 
+    if abort:
+        print("Fatal exception occured, aborting...")
+        sys.exit(1)
```
larskarlitski: Why don't you exit right away when the exception happens?

Please include a hint that the error happened because there's no space left on the device.

tyll: The idea is to still set the status to "error" and upload the logs before exiting. Also the actual exception that mentions that no space is left on the device is printed in line 546.

larskarlitski: Ah, I overlooked that. Thanks!

PR #41, opened 2018-10-05
FMF support

yontalcar:
`standard-inventory-qcow2` searches for FMF in the current working directory. Run `ansible-playbook` in the directory with tests so FMF is picked up.

### DIFF
--- a/test/run-tests
+++ b/test/run-tests
@@ -95,10 +95,12 @@ def run(*argv, env=None, check=True, cwd=None):
     else:
         envrepr = ""
 
+    cwdrepr = "cd {}; ".format(shlex.quote(cwd)) if cwd else ""
+
     if env:
         runenv.update(env)
 
-    print("+ " + envrepr + " ".join(shlex.quote(a) for a in argv))
+    print("+ " + cwdrepr + envrepr + " ".join(shlex.quote(a) for a in argv))
 
     return subprocess.run(
         argv, env=runenv, check=check, stdout=sys.stdout, stderr=sys.stderr, cwd=cwd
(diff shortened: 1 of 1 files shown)


pcahyna:
LGTM, @tyll any idea about the CI error?

tyll:
Never saw this error before. I just restarted the travis CI job in the web UI maybe it is just a glitch.

pcahyna:
It seems so. Can I merge it?

test/run-tests
```diff
@@ -101,7 +101,7 @@ def run(*argv, env=None, check=True):
     print("+ " + envrepr + " ".join(shlex.quote(a) for a in argv))
```
tyll: It might make sense to add the new working directory to the debug output, maybe:

```
if cwd:
    cwdrepr = " cd {} ".format(shlex.quote(cwd))
else:
    cwdrepr = ""
```

and then use 

```
print("+ " + envrepr + cwdrepr + " ".join(shlex.quote(a) for a in argv))
```

tyll:
Looks good to me in general, did you run it locally against the different roles to see if it breaks something?

test/run-tests
tyll: @yontalcar did you see my comment? Do you agree with this change?

tyll:
The code looks good. Before we deploy it, it makes sense to check if this change breaks any existing tests IMHO. @yontalcar can you do this?

yontalcar:
@tyll How should I test it? Ansible playbooks don't depend on `$PWD` of `ansible-playbook`, do they? Anyway, I have ran `test/run-tests --dry-run` on few PRs and it seems to run fine.

tyll:
@yontalcar thank you. You are right, they should not depend on it, just wanted to be safe.

PR #44, opened 2018-10-09
Update token permissions

tyll:


pcahyna:
I don't see any problem with it, but I did not actually try or investigate what is documented here.

PR #48, opened 2019-01-16
some patches for handling rate-limiting on GitHub

yontalcar:


test/run-tests
```diff
@@ -33,6 +34,35 @@ def sighandler_exit(signo, frame):
     sys.exit(0)
 
 
+def handle_httperror(error):
+    """
+    Try to handle HTTPError.
+    Tries to handle transient errors by waiting.
+
+    Return True if error is handled.
```
tyll: Maybe change the docstring to something like `Sleep when a transient error occured, return True if slept and False otherwise`. There does not seem to be more handling other than sleeping therefore this could be explicitly mentioned.

test/run-tests
```diff
@@ -33,6 +34,35 @@ def sighandler_exit(signo, frame):
...
+    """
+
+    # Handle transient server-side errors
+    if error.response.status_code in (500, 502, 503, 504):
+        print(
+            f">>> Server returned {error.response.status_code} {error.response.reason}, waiting..."
+        )
+        time.sleep(600)
```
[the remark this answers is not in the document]
yontalcar: Good idea, though I'm not sure about best way to do it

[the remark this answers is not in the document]
tyll: Since it not a method of an object, I believe the best way is to use a global variable, for example `transient_error_wait_time`, use it with `global transient_error_wait_time` in `handle_transient_httperrors()` and reset it to a default value when no exception occured for `handle_transient_httperrrors()` to handle. The values should be constants.

Eventually, the code should probably be changed to also retry at least the requests to report the status back but I am not sure if they fail often and it is worth the effort now.

tyll:
@yontalcar did you see my review comments?

yontalcar:
@tyll yes, I did, I'm working on it

yontalcar:
@tyll If you have no further comments, I'll squash related commits and merge it

test/run-tests
```diff
@@ -23,16 +23,64 @@ import urllib.parse
 import urllib.request
 
 import requests
+import cachecontrol
 import yaml
 
 HOSTNAME = socket.gethostname()
 
+transient_error_wait_time_default = 30
```
tyll: constants are written in all-caps by convention in Python, see for example `HOSTNAME`.

test/run-tests
```diff
@@ -23,16 +23,64 @@ import urllib.parse
...
+transient_error_wait_time_max = 600
+transient_error_wait_time = transient_error_wait_time_default
```
tyll: The wait time is not a constant but a global variable, so this name is ok.

test/run-tests
```diff
@@ -23,16 +23,64 @@ import urllib.parse
...
+
+comment_cmd_test_all = "[citest]"
```
tyll: These are all constants.

test/run-tests
```diff
@@ -23,16 +23,64 @@ import urllib.parse
...
+comment_cmd_test_pending = "[citest pending]"
+comment_cmd_test_bad = "[citest bad]"
+
+
+def transient_error_wait_time_reset():
```
tyll: IMHO it is better to have the verb at the beginning, i.e. `reset_transient_error_wait_time()`, it is more natural to read.

test/run-tests
```diff
@@ -23,16 +23,64 @@ import urllib.parse
...
+    Sleep when a transient error occured.
+
+    Return True if slept, False otherwise
+    """
+
+    # Handle transient server-side errors
+    if error.response.status_code in (500, 502, 503, 504):
+        global transient_error_wait_time
+        print(
+            ">>> Server returned {} {}, waiting...".format(
```
tyll: please add the wait time to this message, i.e. `waiting {}s...`

PR #49, opened 2019-02-07
Be more defensive when calculating timeout

yontalcar:
* use date from header instead of local time when calculating sleep time
* if calculated timeout is too small (or even negative), use transient_error_wait_time

### DIFF
--- a/test/run-tests
+++ b/test/run-tests
@@ -79,7 +79,10 @@ def handle_transient_httperrors(error):
         and "X-RateLimit-Remaining" in error.response.headers
         and int(error.response.headers["X-RateLimit-Remaining"]) == 0
     ):
-        timeout = int(error.response.headers["X-RateLimit-Reset"]) - time.time()
+        timeout = max(
+            int(error.response.headers["X-RateLimit-Reset"]) - time.time(),
+            transient_error_wait_time.next(),
+        )
         print(f">>> Rate limiting hit, waiting for {int(timeout)} seconds...")
         time.sleep(timeout)
     else:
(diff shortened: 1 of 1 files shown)


test/run-tests
```diff
@@ -32,14 +33,24 @@ COMMENT_CMD_TEST_ALL = "[citest]"
 COMMENT_CMD_TEST_PENDING = "[citest pending]"
 COMMENT_CMD_TEST_BAD = "[citest bad]"
 
-TRANSIENT_ERROR_WAIT_TIME_DEFAULT = 30
-TRANSIENT_ERROR_WAIT_TIME_MAX = 600
-transient_error_wait_time = TRANSIENT_ERROR_WAIT_TIME_DEFAULT
 
+class transient_error_wait_time:
```
tyll: Classues use `CamelCase` by convention in Python, for example `TransientErrorWaitTime`.

test/run-tests
```diff
@@ -56,25 +67,33 @@ def handle_transient_httperrors(error):
...
         and int(error.response.headers["X-RateLimit-Remaining"]) == 0
     ):
-        timeout = int(error.response.headers["X-RateLimit-Reset"]) - time.time()
+        try:
+            now = calendar.timegm(
+                time.strptime(
+                    error.response.headers["Date"], "%a, %d %b %Y %H:%M:%S %Z"
+                )
+            )
+        except:
```
tyll: To please the CI, please use `except ValueError` - also on second thought, this should probably log a warning so that we notice this eventually.

yontalcar: makes sense, will do

PR #50, opened 2019-02-08
Patches from staging branch

yontalcar:
I'd like to receive some comments regarding name config, especially how it will be written in config.json (now it is top level key "test_harness_name")

### DIFF
--- a/test/run-tests
+++ b/test/run-tests
@@ -714,6 +714,16 @@ def handle_task(gh, args, config, task, dry_run=False):
         sys.exit(1)
 
 
+def check_environment():
+    """
+    Check whether the environment is sane.
+    Intent here is to fail early for example if /dev/kvm is not available.
+    """
+    if not os.access("/dev/kvm", os.R_OK | os.W_OK):
+        print("test-harness needs access to /dev/kvm, aborting")
+        sys.exit(1)
+
+
 def main():
     signal.signal(signal.SIGTERM, sighandler_exit)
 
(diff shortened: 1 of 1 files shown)


test/run-tests
```diff
@@ -443,7 +443,7 @@ def get_comment_commands(gh, owner, repo, pullnr):
     return commands
 
 
-def choose_task(gh, repos, images):
+def choose_task(gh, repos, images, config):
```
tyll: In general IMHO the architecture should be changed to for example a `TestHarness` class which has properties like the config, so that the methods like `choose_task()` could just access them via `self.config`.

test/run-tests
```diff
@@ -468,7 +468,11 @@ def choose_task(gh, repos, images):
             commands = get_comment_commands(gh, owner, repo, number)
 
             for image in images:
-                status = statuses.get("linux-system-roles-test/" + image["name"])
+                status_context = "%s/%s" % (
+                    config.get("test_harness_name", "linux-system-roles-test"),
```
tyll: Instead of defining default values at each place where the config is accessed, it makes more sense to have a dictionary with default values and then update those after parsing the config.json. Otherwise the default values need to be kept in sync (for this it should be at least a constant instead of the same string in several places).

tyll: Also I think it could just be `name` in config.json. Since it is the config for test-harness there is no need to add it to the config option as well IMHO.

test/run-tests
```diff
@@ -690,6 +695,16 @@ def handle_task(gh, args, config, task, dry_run=False):
         sys.exit(1)
 
 
+def sanity():
```
tyll: Just noticed, it is better to name methods with a verb, for example `check_environment()`.

PR #54, opened 2019-05-28
Shuffle list of images/repos once

tyll:
To limit the requests to the GitHub API, shuffle the images/repos once
and then handle the first possible task. So avoids making requests to
find all possible tasks to only choose one of them at the end.

### DIFF
--- a/test/run-tests
+++ b/test/run-tests
@@ -487,13 +487,13 @@ def choose_task(gh, repos, images, config):
     Collect tasks from open pull requests (one task for each image
     and each open pull).
 
-    Return one of those tasks at random, so that we reduce the probability of
-    choosing the same one as another instance of this script.
+    Return the first found task. The caller needs to provide shuffled
+    repos/images to reduce the probability that other instances choose the same
+    task.
 
     Returns None if there's nothing to do.
     """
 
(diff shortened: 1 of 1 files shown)


yontalcar:
Unless you are talking about running run-tests with PRs as arguments, I don't see how will this help as tasks are already chosen randomly from PRs×images.

tyll:
> Unless you are talking about running run-tests with PRs as arguments, I don't see how will this help as tasks are already chosen randomly from PRs×images.

Good point, I changed the code to shuffle the images/repos only once when the script starts. This allows to reduce the amount of requests to the GitHub API.

test/run-tests
```diff
@@ -472,13 +472,13 @@ def choose_task(gh, repos, images, config):
...
+    Return the first found task. The caller needs to provide shuffled
+    repos/images to reduce the probability that other instances choose the same
+    task.
 
     Returns None if there's nothing to do.
     """
 
-    tasks = []
     for owner, repo in repos:
         for pull in gh.get(f"repos/{owner}/{repo}/pulls").json():
```
yontalcar: we might want to shuffle pull requests as well

PR #62, opened 2019-12-04
Extend logging capabilities

yontalcar:


test/run-tests
```diff
@@ -622,6 +624,13 @@ def handle_task(gh, args, config, task, dry_run=False):
     status_context = f"{config['name']}/{task.image['name']}"
 
     if not dry_run:
+        logging.info(
+            "Claiming task %s/%s/%d:%s",
```
tyll: Maybe consider adding a `__str__`  method to Tasks, so that you can just use `logging.info("Claiming %s", task)` instead of repeating the details in each log call. `___str___()` could then return something like `task linux-system-roles/network/123:Fedora`

test/run-tests
```diff
@@ -755,6 +779,35 @@ def check_environment():
...
+    if "filename" in log_cfg:
+        if "HOSTNAME" in log_cfg["filename"]:
+            log_cfg["filename"] = log_cfg["filename"].replace("HOSTNAME", HOSTNAME)
+        else:
+            log_cfg["filename"] = log_cfg["filename"] + HOSTNAME
+
+    try:
+        logging.basicConfig(**log_cfg)
+    except ValueError:
+        # invalid logging config, ignore it
```
tyll: This comment is not needed, the warning says this already

test/run-tests
```diff
@@ -755,6 +779,35 @@ def check_environment():
...
+        logging.basicConfig(**log_cfg_default)
+        logging.warning("Invalid logging config", exc_info=True)
```
tyll: The warning could be added to say `using default`.

test/run-tests
```diff
@@ -755,6 +779,35 @@ def check_environment():
         sys.exit(1)
 
 
+def setup_logging(config_logging):
+    log_cfg_default = {
+        "format": "%(asctime)s: %(levelname)s: %(message)s",
+        "datefmt": "%Y-%m-%dT%H:%M:%S",
```
tyll: IMHO it would be better to mention the UTC offset in the time format to make it easier to compare logs from different sources.

test/run-tests
yontalcar: Good idea, will do.

test/run-tests
```diff
@@ -745,6 +768,35 @@ def check_environment():
...
+    }
+
+    log_cfg = log_cfg_default.copy()
+    log_cfg.update(
+        (k, v)
+        for k, v in config_logging.items()
+        if k not in ("stream", "handlers", "force")
```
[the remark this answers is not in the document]
yontalcar: `stream` and `handlers` are Python objects and `force` modifies behaviour of `logging.basicConfig()`; but I was overthinking it – worst case it throws `ValueError`

test/run-tests
```diff
@@ -799,6 +851,9 @@ def main():
         repos = [r.split("/") for r in config.get("repositories", [])]
         random.shuffle(repos)
 
+    if "logging" in config:
```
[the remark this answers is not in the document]
yontalcar: Huh, I just noticed that this conditional is superfluous. So thanks for highlighting this line.

Ok, I'll amend README.

Dockerfile
```diff
@@ -14,6 +14,7 @@ USER tester
 COPY test /test
 
 VOLUME /config /secrets /cache
+VOLUME /log
```
tyll: Is this intentional? It does not seem to be used and if it would be, it needs to be documented.

yontalcar: Hmm… I can mount volume without this entry in Dockerfile so it's on its way out

test/run-tests
```diff
@@ -745,6 +768,35 @@ def check_environment():
...
+        if k not in ("stream", "handlers", "force")
+    )
+
+    # as config may be shared between multiple hosts with shared storage, we
+    # need hostname in log name
+    if "filename" in log_cfg:
+        if "HOSTNAME" in log_cfg["filename"]:
+            log_cfg["filename"] = log_cfg["filename"].replace("HOSTNAME", HOSTNAME)
+        else:
+            log_cfg["filename"] = log_cfg["filename"] + HOSTNAME
```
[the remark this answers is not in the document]
yontalcar: I guess I'll just ignore case w/o `"HOSTNAME"`

test/run-tests
```diff
@@ -100,6 +100,7 @@ def handle_transient_httperrors(error):
             transient_error_wait_time.next(),
         )
 
+        logging.info("Rate limiting hit, waiting for %s seconds", int(timeout))
```
tyll: IMHO this is rather a warning than just an info.

test/run-tests
```diff
@@ -100,6 +100,7 @@ def handle_transient_httperrors(error):
...
         print(f">>> Rate limiting hit, waiting for {int(timeout)} seconds...")
```
tyll: Since this is basically the same message, how about adding a second log handler for level warn that outputs to stdout so that there is not need for a logging message and a print?

yontalcar: OK, I'll have to rework `setup_logging()` though

test/run-tests
```diff
@@ -110,13 +110,12 @@ def handle_transient_httperrors(error):
         except Exception:
             github_msg = ""
         else:
-            print("W: GitHub: " + github_msg)
+            logging.error("GitHub: %s", github_msg)
```
richm: Should this be a warning?  I'm assuming that's what the `W:` meant in the original message

test/run-tests
```diff
@@ -110,13 +110,12 @@ def handle_transient_httperrors(error):
...
 
         if "abuse detection" in github_msg:
             timeout = transient_error_wait_time.next()
-            print(
-                ">>> We have triggered abuse detection, waiting {} s...".format(timeout)
-            )
+            msg = "We have triggered abuse detection, waiting {} s...".format(timeout)
+            logging.warning("%s", msg)
```
richm: `logging.warning("We have triggered abuse detection, waiting %d s...", timeout)`

README.md
```diff
@@ -86,6 +86,15 @@ accessible URL for these results. For example:
   "results": {
     "destination": "user@example.org:public_html/logs",
     "public_url": "https://example.com/logs"
+  },
+  "logging": {
+    "level": "WARNING",
+    "filename": "/var/log/test-harness_HOSTNAME.log",
```
richm: Note that when running in a container, this file will be inside the container, so you will have to do something like `oc exec $pod-name -- cat /var/log/test-harness_HOSTNAME.log` in order to view the contents (or `oc rsh $pod-name` to login to the pod and manually navigate).

richm:
needs rebase, and couple of minor nits, but overall lgtm

richm:
obsoleted by https://github.com/linux-system-roles/test-harness/pull/71

PR #63, opened 2020-01-07
Fix getting statuses

yontalcar:
Currently used API endpoint returns paginated list of all statuses
so old entries may be pushed to second page and further. This may
lead to unnecessary retests when test-harness thinks that test result
is absent.

There is an API endpoint which returns most recent status for each
context. Use this one as we want newest statuses only anyway.

### DIFF
--- a/test/run-tests
+++ b/test/run-tests
@@ -455,14 +455,10 @@ def get_statuses(gh, owner, repo, sha):
     mapping context to its most recent status.
     """
 
-    statuses = {}
-    for status in gh.get(f"repos/{owner}/{repo}/statuses/{sha}").json():
-        # only add the first occurence of a context, since the result is in
-        # reverse chronological order:
-        # https://developer.github.com/v3/repos/statuses/#list-statuses-for-a-specific-ref
-        statuses.setdefault(status["context"], status)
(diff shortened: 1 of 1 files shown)


test/run-tests
```diff
@@ -456,11 +466,21 @@ def get_statuses(gh, owner, repo, sha):
     """
 
     statuses = {}
-    for status in gh.get(f"repos/{owner}/{repo}/statuses/{sha}").json():
-        # only add the first occurence of a context, since the result is in
-        # reverse chronological order:
-        # https://developer.github.com/v3/repos/statuses/#list-statuses-for-a-specific-ref
-        statuses.setdefault(status["context"], status)
+
+    # statuses may be paginated
```
pcahyna: Could you please add a reference to docs about pagination, for those who are not familiar with the concept?

yontalcar:
Huh, [this](https://developer.github.com/v3/repos/statuses/#get-the-combined-status-for-a-specific-ref) already returns most recent status for each context so I guess I'll ditch current commit and use that other API endpoint.

Note that [pagination](https://developer.github.com/v3/#pagination) issue may still show up in some other query returning list.

yontalcar:
Thanks for info. I'll take a look.

PR #69, opened 2020-03-16
Use Ansible for deployment; use BuildConfig for building

richm:
Added a playbook for deploying the CI environment to an
OpenShift cluster.

```
$ ansible-playbook -e test_harness_secrets_dir=$SECRETS_PATH \
    -e test_harness_config_dir=$CONFIG_PATH \
    -e test_harness_scc=$SCC_NAME \
    ansible/openshift-playbook.yml
```

Parameters:
* `test_harness_secrets_dir` - Required - See `SECRETS_PATH`
* `test_harness_config_dir` - Required - See `CONFIG_PATH`
* `test_harness_scc` - Required - See `SCC_NAME`
* `test_harness_namespace` - Default `lsr-test-harness`
* `test_harness_sa` - Default `system:serviceaccount:{{ test_harness_namespace }
}:tester`
* `test_harness_need_node_selector` - Default `true`
* `test_harness_run_as_root` - Default `true`
* `test_harness_node_selector` - Default `{"system-roles-ci": "true"}`
* `test_harness_use_staging` - Default `true`
* `test_harness_use_production` - Default `false`

By default, this will deploy and/or update the staging environment
since`test_harness_use_staging: true`.  It will also change the
DeploymentConfig to use the nodeSelector above, and will configure the
DeploymentConfig to run as root.  If you want to deploy to production, you must
explicitly set `test_harness_use_production` to `true`.

I also added BuildConfigs for the staging and production images so
that we do not have a dependency on `docker.io` or other external
image registry.  I updated the `image` and `imagestream` definitions
to reflect this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,15 @@ all playbooks matching `tests/tests*.yml` (intentionally identical to
 the pattern used in the Fedora Standard Test Interface) of the
 repository against various virtual machines.
 
-To build the container, run
+If using OpenShift, see below.
+
+To build the container for local testing, run
+
+```
+podman build -t linuxsystemroles/test-harness:latest .
+```
+
+If `podman` isn't working, try `docker`:
 
 ```
 sudo docker build -t linuxsystemroles/test-harness:latest .
@@ -117,7 +125,7 @@ In this section we first describe how to setup an OpenShift environment and run
 the CI system on it. Then we give several tips concerning maintenance the
 entire system and backing up important data.
 
-### Preparing OpenShift Environment and Running the CI System on it
+### Preparing OpenShift Environment
 
 In our use case, we create OpenShift project with two deployments. The first
 one will be used for testing and will be based on
@@ -136,9 +144,10 @@ empty directory and put the secrets inside it (files `github-token`, `id_rsa`,
 In what follows, suppose you have set following environment variables:
 
 * `PROJECT_NAME`: A name of OpenShift project.
-* `CONFIG_PATH`: A path to `config.json`.
-* `CONFIG_STAGING_PATH`: A path to `config-staging.json`.
+* `CONFIG_PATH`: A path to directory with `config.json` and `config-staging.json`
 * `SECRETS_PATH`: A path to the directory with secrets.
+* `SCC_NAME`: The name of the SecurityContextConstraints that you will add your
+              ServiceAccount to for privileged operations (e.g. `privileged`)
 
 First, you need to be logged on (replace `[URL]` with a real URL of your
 OpenShift server):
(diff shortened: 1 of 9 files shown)


pcahyna:
Hard to review, it is the first time I see anything like this :-)

richm:
> Hard to review, it is the first time I see anything like this :-)

Understood - so perhaps @yontalcar or @i386x will need to review - but see https://docs.openshift.com/container-platform/3.11/dev_guide/builds/index.html

tyll:
Ideally, the commit message would contain a topic, like:

`Openshift: Add BuildConfigs for production and staging`

or

`Add Openshift BuildConfigs`

see: https://chris.beams.io/posts/git-commit

> This allows us to build the system roles test image inside
> the cluster, so we don't have to depend on pushing the
> image to some external registry.

This information belongs into the commit message IMHO.

> 
> Once deployed, you can edit the build config
> `oc edit bc linux-system-roles-staging`
> 
> The key thing to edit is
> 
> ```yaml
>       source:
>         git:
>           uri: https://github.com/linux-system-roles/test-harness
>           ref: master
> ```
> 
> For example, when the `staging` branch is stable, change it to
> `ref: stable`. If you want to test your local branch in the
> staging environment, use
> 
> ```yaml
>       source:
>         git:
>           uri: https://github.com/myusername/test-harness
>           ref: my-test-branch
> ```
> 
> To do a build, use `oc start-build --follow linux-system-roles-staging`
> Once the build is complete, this should automatically trigger
> a redeployment of the pod(s).

This information should be added to the README.

Hope this is helpful. The yaml structure seem to be fine but I lack the knowledge to check any semantic problems.

richm:
Thanks @tyll .  I have completely revamped the PR.  It didn't make sense to me to have to follow a bunch of manual steps from a README, so I created a playbook.  Now deployment is basically one step (once you are logged into the cluster).  This incorporates the previous commit which added BuildConfigs but goes farther in eliminating the use of `docker.io` images in favor of `ImageStream`s in the cluster.

openshift-playbook.yml
```diff
@@ -0,0 +1,90 @@
+---
+- name: Configure OpenShift deployment for linux-system-roles CI testing
+  hosts: localhost
```
tyll: This should probably be a variable, too.

richm: What shall I name the variable?  This host is the host where I want the playbook to be run, so that it has access to the `oc` command and your current OpenShift authentication context.

create-configmap.yml
```diff
@@ -0,0 +1,26 @@
+---
```
tyll: This is going to drop a lot more files into the repos root directory. It might make sense to move the tasks into a subdir or maybe also the openshift or ansible content.

richm: should I move all of the playbooks/tasks files into `playbooks/`?

openshift-playbook.yml
richm: That is, this host is not the host where I want to install something into.  I suppose this could be something other than `localhost` if you wanted to run ansible on your local workstation, that uploaded the files/scripts to some sort of jump host, which has the `oc` commands and your OpenShift authentication context?  Seems unlikely - I would rather leave this as `localhost` unless there is some use case I am unaware of.

create-configmap.yml
richm: > This is going to drop a lot more files into the repos root directory. It might make sense to move the tasks into a subdir or maybe also the openshift or ansible content.

There is already a lot of stuff in the root directory.  I suppose I could clean up this content - move it into sub-directories as part of this PR, but I think that would be better done in another separate PR.

openshift-playbook.yml
tyll: Ah, I see. Then it does not need to be a variable but the README should mention how to select the cluster for the playbook.

create-configmap.yml
tyll: > should I move all of the playbooks/tasks files into `playbooks/`?

I am not sure how much the other files and the playbook relate, here are possible ideas:

/ansible/tasks/ - for tasks
/ansible/ - for playbooks
/openshift - for OpenShift helpers and files

an alternative could be:

/openshift - for OpenShift tools including the playbook
/openshift/files - for files used by tools and playbook
/openshift/task - for Ansible tasks

> > This is going to drop a lot more files into the repos root directory. It might make sense to move the tasks into a subdir or maybe also the openshift or ansible content.
> 
> There is already a lot of stuff in the root directory. I suppose I could clean up this content - move it into sub-directories as part of this PR, but I think that would be better done in another separate PR.

I guess this depends on how much the files/tools depend on each other. If it is not much extra work to do the extra PR, it makes reviewing things and moving forward easier.

richm: > > should I move all of the playbooks/tasks files into `playbooks/`?
> 
> I am not sure how much the other files and the playbook relate, here are possible ideas:
> 
> /ansible/tasks/ - for tasks
> /ansible/ - for playbooks

Done.

> /openshift - for OpenShift helpers and files

I don't want to move these because I don't want to disrupt the workflow of other users of this repo.  See the `README.md` for the directions for the old workflow that doesn't use ansible.

> 
> an alternative could be:
> 
> /openshift - for OpenShift tools including the playbook
> /openshift/files - for files used by tools and playbook
> /openshift/task - for Ansible tasks

See above.

> 
> > > This is going to drop a lot more files into the repos root directory. It might make sense to move the tasks into a subdir or maybe also the openshift or ansible content.
> > 
> > 
> > There is already a lot of stuff in the root directory. I suppose I could clean up this content - move it into sub-directories as part of this PR, but I think that would be better done in another separate PR.
> 
> I guess this depends on how much the files/tools depend on each other. If it is not much extra work to do the extra PR, it makes reviewing things and moving forward easier.

* I don't want to disrupt the existing workflow of other users of this repo
* I don't want to have too much unrelated stuff in this PR

README.md
```diff
@@ -147,7 +156,42 @@ OpenShift server):
...
 ```
 
-Now, you can create a fresh OpenShift project:
+### Installing CI System in OpenShift using Ansible
+
+NOTE: Ansible will be running all of the commands on `localhost`, instead of
+executing the tasks remotely in the cluster.  That is, the Ansible playbook
+will be using the `oc` OpenShift command line tool, and the python OpenShift
+API library, to execute API calls in the OpenShift cluster to provision and
+configure the resources in the OpenShift cluster.
```
tyll: This sounds good. Maybe just move the login command here and to the other section for more visibility.

PR #74, opened 2020-04-01
use fedora 31 image for test pod

richm:


richm:
> +1 CentOS 8 might also be a viable alternative. Did you test this in staging already?

https://github.com/linux-system-roles/timesync/pull/67 - the `linux-system-roles-test-staging/fedora-30` is in progress

richm:
> > +1 CentOS 8 might also be a viable alternative.

@tyll Looks like it won't be straightforward to build or run - just using centos:8 in FROM fails to build:
```
STEP 2: RUN dnf update -y && dnf install -y     ansible     dumb-init     git     python3-requests     python3-CacheControl     standard-test-roles && dnf clean all
...
No match for argument: ansible
No match for argument: dumb-init
No match for argument: python3-CacheControl
No match for argument: standard-test-roles
Error: Unable to find a match: ansible dumb-init python3-CacheControl standard-test-roles
Error: error building at STEP "RUN dnf update -y && dnf install -y     ansible     dumb-init     git     python3-requests     python3-CacheControl     standard-test-roles && dnf clean all": error while running runtime: exit status 1
```
So, looks like someone would need to put in some time to track down all of these packages for centos8, assuming they or similar packages exist . . .

What would be the advantage of using centos8?  What is the disadvantage of using fedora, other than having to upgrade to the latest version of fedora every 6 months or otherwise periodically?

PR #81, opened 2020-05-19
Dockerfile: Update from Fedora 31 to Fedora 32

jharuda:
It updates Ansible version.


tyll:
Fedora 32 is using Python 3.8, so it might also make sense to update tox to test against Python 3.8.

pcahyna:
> Fedora 32 is using Python 3.8, so it might also make sense to update tox to test against Python 3.8.

Thanks, I see this led to #82 .

PR #82, opened 2020-05-19
fix flake8 error; use python 3.8 for tox/travis

richm:


test/run-tests
```diff
@@ -679,8 +679,8 @@ def handle_task(gh, args, config, task, dry_run=False):
             status = statuses.get(status_context)
             if status["description"] != description:
                 logging.info(
-                    f"Skip: another instance is working on this task: "
-                    + status["description"]
+                    "Skip: another instance is working on this task: "
+                    f"{status['description']}"
```
pcahyna, line 683: I was surprised that this works, but indeed it does: https://docs.python.org/3/reference/lexical_analysis.html#string-literal-concatenation "formatted string literals may be concatenated with plain string literals".

PR #80, opened 2020-05-05
use master branch for staging; use production branch for production

richm:
Also upgrades ci to use py37 and ubuntu bionic

PR #86, opened 2020-06-22
Advise to do fast forward merges for PRs and production

pcahyna:
With instructions and examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,32 @@ It is also possible to whitelist all commits in a PR with the `needs-ci` tag.
 However, this should be only used with trusted submitters since they can still
 change the commits in the PR.
 
+## Git workflow
+
+1. Merge pull requests to master without merge commits to keep a less
+   cluttered history.
+1. Make sure that `production` is always an ancestor of `master` (or
+   equivalent to `master`). Having `production` as a side branch and
+   continuously merging from `master` would lead to a complicated
+   history and also would not ensure that `production` is in a state
+   that was previously tested on `master`.
+
+Ad 1: when you want to merge a PR, fetch the branch with the PR,
+checkout `master` and merge the branch using `git merge` with
+`--ff-only`. This will refuse to merge the branch if not up-to-date,
+in this case, rebase it first to `master`.
+
+Example for PR #85: if you have the branch to be merged
+(`ukulekek-pre-flight-checks-openshift`) checked out aready, do
+```sh
+git checkout master
+git merge --ff-only ukulekek-pre-flight-checks-openshift
+git push origin master
+```
+
+Ad 2: When updating `production` to `master`, reset it to `master`, or
+fast-forward it to `master` (using `merge` with `--ff-only`).
+
 ## Deployment Workflow
 
 To make changes to the CI environment, the changes should be first committed


README.md
```diff
@@ -16,6 +16,32 @@ It is also possible to whitelist all commits in a PR with the `needs-ci` tag.
 However, this should be only used with trusted submitters since they can still
 change the commits in the PR.
 
+## Git workflow
+
+1. Merge pull requests to master without merge commits to keep a less
```
tyll, line 21: There are branch protection rules in GitHub, I enabled:

"Require linear history
Prevent merge commits from being pushed to matching branches."

This should take care of this as far as I understand and also allow to use the web interface instead of git commands.

tyll: Also adjusted the setting for the merge button to only allow rebase. The branch protection rule should probably also be added for `production`.

tyll:
Btw. I guess it would also make sense to rename master to staging, so there are only staging and production branches with less ambiguity for the master branch.

richm:
> Btw. I guess it would also make sense to rename master to staging, so there are only staging and production branches with less ambiguity for the master branch.

The team thought it was clear that `staging` == development == `master` branch, and that disambiguation was only needed for `production`.

tyll:
> > Btw. I guess it would also make sense to rename master to staging, so there are only staging and production branches with less ambiguity for the master branch.
> 
> The team thought it was clear that `staging` == development == `master` branch, and that disambiguation was only needed for `production`.

This describes my observation: Since this defines the meaning of "master" only by the presence of the other branch as in 'if there is another branch called "production", this means that "master" is probably not production but development' but it is not clear by looking at master by itself.

richm:
So can we merge this PR?  lgtm

jharuda:
LGTM

README.md
pcahyna: @tyll, thanks, will it perform a fast-forward merge or do a rebase resulting in a new commit?

tyll: should be a fast-forward merge

pcahyna:
> > > Btw. I guess it would also make sense to rename master to staging, so there are only staging and production branches with less ambiguity for the master branch.
> > 
> > 
> > The team thought it was clear that `staging` == development == `master` branch, and that disambiguation was only needed for `production`.
> 
> This describes my observation: Since this defines the meaning of "master" only by the presence of the other branch as in 'if there is another branch called "production", this means that "master" is probably not production but development' but it is not clear by looking at master by itself.

@jharuda @xjezda00 @ukulekek what is your usual naming of branches, are you typically using `staging` and `production` and avoid `master`?

jharuda:
@pcahyna I usually use **master (==production meaning)** and **staging (testing)**  branches or **production** and **staging (testing)** branches style. So at this point I agree with Tyll. But in our use-case there is no need for renaming branches because Test-harness user still needs to read the documentation and after that he knows that master is staging (development) branch.

richm:
> @pcahyna I usually use **staging (testing) -> master (==production meaning)** branches or **staging (testing) -> production** branches style.

Which is what we used to do until it was suggested that instead we use master branch==staging environment and production branch==production environment - https://github.com/linux-system-roles/test-harness/pull/80 . . .

> So at this point I agree with Tyll. But in our use-case there is no need to rename branches because Test-harness user still needs to read the documentation and after that he knows that master is staging (development) branch.

I really don't care what we use at this point - I don't think there is a "One True Way To Do It" that everyone (now, and in the future) will agree on - so we just need to pick something reasonable and move on.

tyll:
> > So at this point I agree with Tyll. But in our use-case there is no need to rename branches because Test-harness user still needs to read the documentation and after that he knows that master is staging (development) branch.
> 
> I really don't care what we use at this point - I don't think there is a "One True Way To Do It" that everyone (now, and in the future) will agree on - so we just need to pick something reasonable and move on.

Use staging/production makes it explicit/unambiguous and does not rely on anyone needing to remember which special meaning "master" has, therefore I am in favor of this.

richm:
I think this is ready to be merged (after rebase)

PR #87, opened 2020-07-01
Add support for using ansible 2.7/python 2/centos7 as controller

richm:
This adds a container/pod for running tests using ansible 2.7 and
python 2 on centos7 as a controller node, in addition to the current
one which uses ansible 2.9/python3/Fedora 32.

The container is run in a new pod/deployment called
linux-system-roles-centos7 (and -staging).  This runs alongside
the current one.  The new one uses the same configuration as
the old one.

The status in the PR now adds the ansible version like this:

linux-system-roles-test/centos-7/ansible-2.7
or
linux-system-roles-test/fedora-31/ansible-2.9

This will be used for *all* statuses going forward.  That is,
there will be no more `linux-system-roles-test/fedora-31` - all
statuses, even for the current ansible, will include the ansible
version string.

The `run-tests` script running in the pod can now be controlled
by setting environment variables.  See the README.md for details.

### DIFF
--- a/Dockerfile.centos7
+++ b/Dockerfile.centos7
@@ -1,0 +1,23 @@
+FROM centos:7
+
+RUN yum update -y && PKGS="centos-release-ansible-27 centos-release-scl-rh" && \
+    yum install -y $PKGS && rpm -V $PKGS && \
+    PKGS="ansible epel-release" && yum install -y $PKGS && rpm -V $PKGS && \
+    PKGS="rh-git218 python3-pip standard-test-roles python2-fmf seabios-bin" && \
+    yum -y install $PKGS && rpm -V $PKGS && \
+    yum clean all && \
+    pip3 install cachecontrol
+
+RUN curl -L -o /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_x86_64
+RUN chmod +x /usr/local/bin/dumb-init
+
+RUN useradd -m tester
+USER tester
+
+COPY test /test
+
+VOLUME /config /secrets /cache
+
+WORKDIR /home/tester
+ENTRYPOINT ["/usr/local/bin/dumb-init", "--"]
+CMD ["scl", "enable", "rh-git218", "/test/run-tests"]
--- a/README.md
+++ b/README.md
@@ -21,7 +21,9 @@ change the commits in the PR.
 To make changes to the CI environment, the changes should be first committed
 to the `master` branch, and deployed to the `staging` environment.  Once they
 are confirmed to be working, the changes should be merged to the `production`
-branch and deployed to the `production` environment.
+branch and deployed to the `production` environment.  For the centos7 based
+tests, use `linux-system-roles-centos7` for production and use
+`linux-system-roles-centos7-staging` for staging.
 
 Steps:
 
@@ -106,6 +108,7 @@ accessible URL for these results. For example:
       "name": "fedora-28",
       "source": "https://download.fedoraproject.org/pub/fedora/linux/releases/28/Cloud/x86_64/images/Fedora-Cloud-Base-28-1.1.x86_64.qcow2",
       "upload_results": true,
+      "min_ansible_version": "2.7",
       "setup": [
         {
           "name": "Setup",
@@ -149,6 +152,20 @@ The `setup` key contains either a list of Ansible plays which will be saved as
 a playbook and executed before the test run, or a single shell command which
 will be executed using the Ansible `raw` module (so the two examples above are
 exactly equivalent).
+
+Use the `min_ansible_version` key to indicate the minimum version of ansible
+that can be used.  For example, if you need to use ansible version 2.8 or
+later to manage Fedora 32 hosts, use something like this:
+```json
+    {
+      "name": "fedora-32",
+      "source": "http://download.fedoraproject.org/pub/fedora/linux/releases/32/Cloud/x86_64/images/Fedora-Cloud-Base-32-1.6.x86_64.qcow2",
+      "upload_results": true,
+      "min_ansible_version": "2.8"
+    },
+```
+This means the CI environment will run tests using only ansible 2.8 or later
+to manage Fedora 32 hosts.
 
 The container needs a `/secrets` mount, which must contain these files:
 
@@ -439,4 +456,34 @@ NAME                                  READY     STATUS       RESTARTS   AGE
 linux-system-roles-12-ddtrs           1/1       Running      57         55d
 ...
 $ oc exec linux-system-roles-12-ddtrs -- cat /var/log/test-harness_linux-system-roles-12-ddtrs.log
+```
+
+## run-tests Command Line Args and Environment Variables
+
+The `run-tests` script can be controlled via command line arguments,
+environment variables, and the config.json config file.  When running in a pod
+in a Kubernetes/OpenShift environment, the easiest way to customize the pod is
+via environment variables.  The environment variables correspond to the
+command line arguments to `run-tests`. In general, the name of the environment
+variable is the name of the command line argument, in all upper case, with `_`
+instead of `-`, and with the prefix `TEST_HARNESS_`.  For example, the command
+line option `--use-images` can be set via the env. var.
+`TEST_HARNESS_USE_IMAGES`.  See `run-tests` usage for the complete list.
+
+The precedence is:
+* command line args are highest precedence
+* environment variables are next highest
+* settings in the config.json file are next highest
+* defaults are lowest precedence
+
+For example, in a running OpenShift cluster, if you wanted to use a custom
+variant in the string used for the PR status:
+```
+oc set env dc/linux-system-roles-staging TEST_HARNESS_VARIANT=my_custom_name
+```
+This will cause the staging pods to be redeployed and use `my_custom_name` for
+the PR status reporting string e.g. `centos-7/ansible-29 (my_custom_name)`.
+To reset:
+```
+oc set env dc/linux-system-roles-staging TEST_HARNESS_VARIANT-
 ```
--- a/test/run-tests
+++ b/test/run-tests
@@ -28,6 +28,11 @@ import cachecontrol
 import cachecontrol.heuristics
 import yaml
 
+# this is what Ansible uses for version comparison
+# use LooseVersion to handle alphas, betas, other pre-release versions
+from distutils.version import LooseVersion
+from distutils.util import strtobool
+
 HOSTNAME = socket.gethostname()
 
 COMMENT_CMD_TEST_ALL = "[citest]"
@@ -199,7 +204,7 @@ class DontCacheStatuses(cachecontrol.heuristics.BaseHeuristic):
         pass
 
 
-def run(*argv, env=None, check=True, cwd=None):
+def run(*argv, env=None, check=True, cwd=None, return_stdout=False):
     """
     Small wrapper around subprocess.run(), which prints the command to be
     executed and raises an exception by default.
@@ -216,9 +221,18 @@ def run(*argv, env=None, check=True, cwd=None):
         runenv.update(env)
 
     print("+ " + cwdrepr + envrepr + " ".join(shlex.quote(a) for a in argv))
+    stdout_val = sys.stdout
+    if return_stdout:
+        stdout_val = subprocess.PIPE
 
     return subprocess.run(
-        argv, env=runenv, check=check, stdout=sys.stdout, stderr=sys.stderr, cwd=cwd
+        argv,
+        env=runenv,
+        check=check,
+        stdout=stdout_val,
+        stderr=sys.stderr,
+        cwd=cwd,
+        encoding="utf-8",
     )
 
 
@@ -504,7 +518,15 @@ def get_comment_commands(gh, owner, repo, pullnr):
     return commands
 
 
-def choose_task(gh, repos, images, config):
+def get_status_context(variant, image_name, ansible_id):
+    if variant:
+        suffix = f" ({variant})"
+    else:
+        suffix = ""
+    return f"{image_name}/{ansible_id}{suffix}"
+
+
+def choose_task(gh, repos, images, config, ansible_id, variant):
     """
     Collect tasks from open pull requests (one task for each image
     and each open pull).
@@ -532,7 +554,7 @@ def choose_task(gh, repos, images, config):
             commands = get_comment_commands(gh, owner, repo, number)
 
             for image in images:
-                status_context = f"{config['name']}/{image['name']}"
+                status_context = get_status_context(variant, image["name"], ansible_id)
                 status = statuses.get(status_context)
 
                 if check_commit_needs_testing(status, commands):
@@ -636,7 +658,7 @@ def make_html(source_file):
     return html_file
 
 
-def handle_task(gh, args, config, task, dry_run=False):
+def handle_task(gh, args, config, task, ansible_id, dry_run=False):
     """ Process a task """
     title = f"{HOSTNAME}: {task.owner}/{task.repo}: pull #{task.pull} "
     title += f'({task.head[:7]}) on {task.image["name"]}'
@@ -650,7 +672,7 @@ def handle_task(gh, args, config, task, dry_run=False):
     target_url = None
     state = None
 
-    status_context = f"{config['name']}/{task.image['name']}"
+    status_context = get_status_context(args.variant, task.image["name"], ansible_id)
 
     if not dry_run:
         logging.info("Claiming %s", task)
@@ -750,7 +772,8 @@ def handle_task(gh, args, config, task, dry_run=False):
                     description += ": Error uploading results"
 
             # FIXME: workdir might be kept when python crashes
-            shutil.rmtree(workdir)
+            if not args.keep_results:
+                shutil.rmtree(workdir)
 
     finally:
         duration = (datetime.datetime.utcnow() - start_time).seconds
@@ -827,45 +850,86 @@ def setup_logging(config_logging):
         logging.warning("Invalid logging config, using default", exc_info=True)
 
 
+def get_ansible_version():
+    """determine the version of Ansible"""
+    rs = run("ansible", "--version", return_stdout=True)
+    return rs.stdout.split()[1]
+
+
 def main():
     signal.signal(signal.SIGTERM, sighandler_exit)
     print("Starting at {}".format(time.asctime()))
 
     parser = argparse.ArgumentParser()
-    parser.add_argument("--secrets", default="/secrets", help="Directory with secrets")
     parser.add_argument(
-        "--config", default="/config", help="Directory with config.json"
+        "--secrets",
+        default=os.environ.get("TEST_HARNESS_SECRETS", "/secrets"),
+        help="Directory with secrets",
     )
     parser.add_argument(
-        "--cache", default="/cache", help="Directory for caching VM images"
+        "--config",
+        default=os.environ.get("TEST_HARNESS_CONFIG", "/config"),
+        help="Directory with config.json",
+    )
+    parser.add_argument(
+        "--cache",
+        default=os.environ.get("TEST_HARNESS_CACHE", "/cache"),
+        help="Directory for caching VM images",
     )
     parser.add_argument(
         "--inventory",
+        default=os.environ.get(
+            "TEST_HARNESS_INVENTORY",
+            "/usr/share/ansible/inventory/standard-inventory-qcow2",
+        ),
         help="Inventory to use for VMs",
-        default="/usr/share/ansible/inventory/standard-inventory-qcow2",
     )
+    default_pull_request = os.environ.get("TEST_HARNESS_PULL_REQUEST", None)
+    if default_pull_request:
+        default_pull_request = default_pull_request.split(",")
     parser.add_argument(
         "pull_request",
         nargs="*",
-        default=None,
+        default=default_pull_request,
         help="Pull requests to test. Example: " "linux-system-roles/network/1",
     )
     parser.add_argument(
         "--use-images",
-        default="*",
-        help="Test pull request only against images matching this pattern",
+        default=os.environ.get("TEST_HARNESS_USE_IMAGES", "*"),
+        help=(
+            "Test pull request only against images matching these patterns.  "
+            "This is a comma delimited list of fnmatch patterns which will be "
+            "applied to each image name and source.  If any of the patterns "
+            "match the name or source, the image will be included in testing."
+        ),
     )
     parser.add_argument(
         "--dry-run",
-        help="Do not update pull request status or upload artifacts",
-        default=False,
+        default=bool(strtobool(os.environ.get("TEST_HARNESS_DRY_RUN", "False"))),
         action="store_true",
+        help="Do not update pull request status or upload artifacts",
+    )
+    parser.add_argument(
+        "--keep-results",
+        default=bool(strtobool(os.environ.get("TEST_HARNESS_KEEP_RESULTS", "False"))),
+        action="store_true",
+        help="For debugging - do not remove the temp results directory",
+    )
+    parser.add_argument(
+        "--variant",
+        default=os.environ.get("TEST_HARNESS_VARIANT", ""),
+        help="Use this variant for PR test status e.g. staging",
+    )
+    parser.add_argument(
+        "--repositories",
+        default=os.environ.get("TEST_HARNESS_REPOSITORIES", None),
+        help="Comma delimited list of repositories to override config",
     )
 
     args = parser.parse_args()
 
     # default values for config
-    config = {"repositories": [], "images": [], "name": "linux-system-roles-test"}
+    config = {"repositories": [], "images": []}
 
     with open(args.config + "/config.json") as configfile:
         config.update(json.load(configfile))
@@ -874,10 +938,62 @@ def main():
         repos = [r.split("/") for r in config.get("repositories", [])]
         random.shuffle(repos)
 
+    # this is used in PR status - context name
+    if not args.variant:
+        if "variant" not in config:
+            # legacy support for old configs that still use "name"
+            if config.get("name") == "linux-system-roles-test-staging":
+                args.variant = "staging"
+        else:
+            args.variant = config["variant"]
+
+    if args.repositories:
+        repos = [r.split("/") for r in args.repositories.split(",")]
     setup_logging(config.get("logging", {}))
 
     check_environment()
 
+    ansible_version = LooseVersion(get_ansible_version())
+    logging.info("Using Ansible version {}".format(ansible_version))
+    # this is used in PR status - context name
+    ansible_id = f"ansible-{ansible_version.version[0]}.{ansible_version.version[1]}"
+
+    image_patterns = args.use_images.split(",")
+    # copy all images to test_images . . .
+    test_images = dict([(image["name"], image) for image in images])
+    # . . . then remove the ones that do not match the criteria
+    for image in images:
+        min_ansible_version = LooseVersion(image.get("min_ansible_version", "0"))
+        if ansible_version < min_ansible_version:
+            if image["name"] in test_images:
+                logging.debug(
+                    f"Image {image['name']} will not be used for testing because "
+                    f"the minimum Ansible version {min_ansible_version} required "
+                    f"by the image is greater than the version {ansible_version} "
+                    "of Ansible used by the test."
+                )
+                del test_images[image["name"]]
+            continue
+        pattern_matched = False
+        for pattern in image_patterns:
+            if fnmatch.fnmatch(image["name"], pattern) or fnmatch.fnmatch(
+                image["source"], pattern
+            ):
+                pattern_matched = True
+                break
+        if not pattern_matched:
+            if image["name"] in test_images:
+                logging.debug(
+                    f"Image {image['name']} will not be used for testing "
+                    "because neither the name nor the source match any "
+                    f"of the image patterns in {image_patterns}"
+                )
+                del test_images[image["name"]]
+
+    logging.info(
+        f"Will test with the following image names: {list(test_images.keys())}"
+    )
+    test_images = list(test_images.values())
     if args.dry_run:
         token = ""
     else:
@@ -926,23 +1042,13 @@ def main():
             head = pull["head"]["sha"]
         number = pull["number"]
 
-        image_patterns = args.use_images.split(",")
-        test_images = []
-        for pattern in image_patterns:
-            for image in images:
-                if fnmatch.fnmatch(image["name"], pattern) or fnmatch.fnmatch(
-                    image["source"], pattern
-                ):
-                    if image not in test_images:
-                        test_images.append(image)
-
         for image in test_images:
             task = Task(owner, repo, number, head, image)
-            handle_task(gh, args, config, task, args.dry_run)
+            handle_task(gh, args, config, task, ansible_id, args.dry_run)
 
     while not args.pull_request:
         try:
-            task = choose_task(gh, repos, images, config)
+            task = choose_task(gh, repos, test_images, config, ansible_id, args.variant)
             if not task:
                 if not printed_waiting:
                     logging.info(">>> No tasks. Waiting.")
@@ -952,7 +1058,7 @@ def main():
                 time.sleep(600)
                 continue
 
-            handle_task(gh, args, config, task, args.dry_run)
+            handle_task(gh, args, config, task, ansible_id, args.dry_run)
         except requests.exceptions.HTTPError as err:
             if not handle_transient_httperrors(err):
                 raise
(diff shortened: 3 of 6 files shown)


test/run-tests
```diff
@@ -216,9 +221,13 @@ def run(*argv, env=None, check=True, cwd=None):
...
 
     print("+ " + cwdrepr + envrepr + " ".join(shlex.quote(a) for a in argv))
+    stdout_val = sys.stdout
+    if return_stdout:
+        stdout_val = subprocess.PIPE
 
     return subprocess.run(
-        argv, env=runenv, check=check, stdout=sys.stdout, stderr=sys.stderr, cwd=cwd
+        argv, env=runenv, check=check, stdout=stdout_val, stderr=sys.stderr, cwd=cwd,
+        text=text
```
richm, line 230: This change was made so that I could use the existing `run` method to get the output of `ansible --version` in order to determine the version of ansible used in the environment.

test/run-tests
```diff
@@ -280,7 +291,16 @@ class checkout_repository:  # pylint: disable=invalid-name
...
-        run("git", "-C", self.dir.name, "checkout", "--quiet", "FETCH_HEAD")
+        run(
+            "git",
+            "--git-dir",
+            self.dir.name + "/.git",
+            "--work-tree",
+            self.dir.name,
+            "checkout",
+            "--quiet",
+            "FETCH_HEAD",
```
richm, line 302: These changes were made so that I could use `git` 1.8 on centos7 - the `-C` argument was added in git 2.0.

test/run-tests
```diff
@@ -532,7 +552,7 @@ def choose_task(gh, repos, images, config):
             commands = get_comment_commands(gh, owner, repo, number)
 
             for image in images:
-                status_context = f"{config['name']}/{image['name']}"
+                status_context = f"{config['name']}/{image['name']}/{ansible_ver_str}"
```
richm, line 555: The `status_context` is the string used in the PR to denote the status e.g. `linux-system-roles-test-staging/centos-7/ansible-2.9` - now the ansible version is added to every status.

test/run-tests
```diff
@@ -867,17 +894,63 @@ def main():
...
+    ansible_version = LooseVersion(get_ansible_version())
+    logging.info("Using Ansible version {}".format(ansible_version))
+    # this is used in PR status - context name
+    ansible_ver_str = (
+        f"ansible-{ansible_version.version[0]}.{ansible_version.version[1]}"
+    )
+
+    image_patterns = args.use_images.split(",")
+    test_images = []
+    for pattern in image_patterns:
```
richm, line 931: This code was refactored so that we could apply the ansible version checking and the image pattern matching once.

richm:
This PR also adds support for a `min_ansible_version` field in the `image` in the config.  We will need to change the config like this:
```json
  "images": [
    {
      "name": "fedora-32",
      "source": "http://download.fedoraproject.org/pub/fedora/linux/releases/32/Cloud/x86_64/images/Fedora-Cloud-Base-32-1.6.x86_64.qcow2",
      "upload_results": false,
      "min_ansible_version": "2.8"
    },
...
```
This means we will only do testing on Fedora 32 with ansible version 2.8 or later i.e. this will be skipped when testing with ansible 2.7.  If no `min_ansible_version` is specified, it will fall back to the `DEFAULT_MIN_ANSIBLE_VERSION = "2.7"`.  That is, by default, all tests will run on all platforms and all versions of ansible we support.

test/run-tests
```diff
@@ -216,9 +221,18 @@ def run(*argv, env=None, check=True, cwd=None):
...
 
     return subprocess.run(
-        argv, env=runenv, check=check, stdout=sys.stdout, stderr=sys.stderr, cwd=cwd
+        argv,
+        env=runenv,
+        check=check,
+        stdout=stdout_val,
+        stderr=sys.stderr,
+        cwd=cwd,
+        text=text,
```
richm, line 235: this change was made because python `black` did not like the old formatting . . .

test/run-tests
pcahyna: Note that there is a newer git available in Software Collections.

pcahyna: https://wiki.centos.org/AdditionalResources/Repositories/SCL . Note that https://wiki.centos.org/SpecialInterestGroup/SCLo/CollectionsList is obsolete, I see rh-git218-git in Collections.

test/run-tests
pcahyna: Will GitHub show such a long status properly?

test/run-tests
richm: I will see if I can use scl git 2 and revert this change

test/run-tests
richm: Yes, mostly - see https://github.com/linux-system-roles/kernel_settings/pull/20
```
 linux-system-roles-test-staging/fedora-32/ansible-2.9 Pending — localhost.localdomain@2020-07-01 14:35:38.813840 
```
In the UI, the timestamp is truncated, but it is also truncated for some of the other longer statuses

If length is a concern, note that most of the length of the status string comes from "linux-system-roles-test-" - so perhaps we could abbreviate that e.g. "lsr-test-" or ???

nhosoi: > If length is a concern, note that most of the length of the status string comes from "linux-system-roles-test-" - so perhaps we could abbreviate that e.g. "lsr-test-" or ???

+1

tyll: Since it being a status it is probably clear that it is a test run, so how about `system-role/ansible-2.9/fedora-32`. or `sysrole/ansible-2.9/fedora-32`? Actually, once could argue that `ansible-2.9/fedora-32` might be enough as well and maybe `ansible-2.9/fedora-32 (staging)` for the staging instance.

richm: ok - I will make this change in a separate PR

richm: > ok - I will make this change in a separate PR

Or is it ok to make this change in this PR?  I must change the status_context string in this PR in order for centos7 testing to work - maybe I should go ahead and drop the "linux-system-roles-test-" too?

nhosoi: I'd like to double-check... If there's no word after linux-system-roles-test like this, how it's going to be displayed?
` linux-system-roles-test/rhel-x — linux-system-roles-39-s828j@2020-06-26 21:19:53.503386`

tyll: > > ok - I will make this change in a separate PR
> 
> Or is it ok to make this change in this PR? I must change the status_context string in this PR in order for centos7 testing to work - maybe I should go ahead and drop the "linux-system-roles-test-" too?

Since you adjust the string it seems fine to me to drop the prefix, too.

tyll: > I'd like to double-check... If there's no word after linux-system-roles-test like this, how it's going to be displayed?
> ` linux-system-roles-test/rhel-x — linux-system-roles-39-s828j@2020-06-26 21:19:53.503386`

This would be `rhel-x/ansible-xy — linux-system-roles-39-s828j@2020-06-26 21:19:53.503386`

nhosoi: Thanks. So, `staging` is rather an exception. :)
+1

richm: and staging would be `rhel-x/ansible-x.y (staging) -- linux-system-roles-39-s828j....`

richm: Here is a real example from a real PR: https://github.com/linux-system-roles/kernel_settings/pull/20
`centos-8/ansible-2.9 (staging) Pending — localhost.localdomain@2020-07-01 18:19:15.258999`

pcahyna: @tyll 
> Actually, once could argue that ansible-2.9/fedora-32 might be enough as well and maybe ansible-2.9/fedora-32 (staging) for the staging instance.

GitHub does not seem to sort the statuses (unfortunately), so whether to use `staging/image/ansible` or `image/ansible (staging)` to indicate the instance is more a matter of taste. Note though that this string is not just a UI/display string. It is used by the test-harness to identify what runs are pending or already completed.

test/run-tests
richm: fixed in https://github.com/linux-system-roles/test-harness/pull/87/commits/f6a2c4331b324de0f736ab11f9ce451f459f9468

Also note - for some reason, this causes the use of py36 which made another change necessary - https://github.com/linux-system-roles/test-harness/pull/87/commits/548311ce89fb1ea8d17fba8063681e336b7a1c18

test/run-tests
richm: fixed in https://github.com/linux-system-roles/test-harness/pull/87/commits/a44492204e08443e57774e011be91e3436e71f77 - see https://github.com/linux-system-roles/kernel_settings/pull/20 for examples of what these new status strings look like

test/run-tests
```diff
@@ -529,6 +518,13 @@ def get_comment_commands(gh, owner, repo, pullnr):
     return commands
 
 
+def get_status_context(config_name, image_name, ansible_ver_str):
+    suffix = ""
+    if config_name.endswith("-staging"):
+        suffix = " (staging)"
```
pcahyna, line 524: I believe the instance name should be configurable instead of having only one option as a special case. There will be other instances than production and staging - if one is developing a change in test-harness or an alternative to it, one typically wants to test it before even pushing it to staging. There are many examples of this already where some PRs have results from some special CI runs. We may also have two different development branches running in parallel at the same time, like the qemu one and openstack one.

richm: ok - how about this https://github.com/linux-system-roles/test-harness/pull/87/commits/5e29e183f524819b1311167deffc61968e33ea7e

README.md
```diff
@@ -150,6 +152,20 @@ a playbook and executed before the test run, or a single shell command which
...
+to manage Fedora 32 hosts, use something like this:
+```json
+    {
+      "name": "fedora-32",
+      "source": "http://download.fedoraproject.org/pub/fedora/linux/releases/32/Cloud/x86_64/images/Fedora-Cloud-Base-32-1.6.x86_64.qcow2",
+      "upload_results": false,
+      "min_ansible_version": "2.8"
+    },
+```
+This means the CI environment will not run tests using ansible 2.7 to
```
tyll, line 166:
```suggestion
This means the CI environment will not run tests using ansible 2.7 or earlier to
```

test/run-tests
```diff
@@ -867,17 +899,63 @@ def main():
...
+    ansible_ver_str = (
+        f"ansible-{ansible_version.version[0]}.{ansible_version.version[1]}"
+    )
+
+    image_patterns = args.use_images.split(",")
+    test_images = []
+    for pattern in image_patterns:
+        for image in images:
+            min_ansible_version = image.get(
+                "min_ansible_version", DEFAULT_MIN_ANSIBLE_VERSION
```
tyll, line 943: Why is the default "2.7" and not just "0", here?

pcahyna: Makes sense to have 0 - of course the minimum supported version is 2.7 now, but hardcoding it will leave future generations wondering whether they should bump it every time an Ansible version gets out of support.

richm: fixed in https://github.com/linux-system-roles/test-harness/pull/87/commits/b41464cbac23050b63fb78c106256e56028731a6

README.md
richm: fixed in https://github.com/linux-system-roles/test-harness/pull/87/commits/e08a9eff0dd3e7052bcbad465dfdc921caf1464f

Dockerfile.centos7
```diff
@@ -0,0 +1,20 @@
+FROM centos:7
+
+RUN yum update -y && yum install -y centos-release-ansible-27 centos-release-scl-rh && \
+    yum -y install ansible && \
+    yum -y install epel-release && \
```
tyll, line 5: I guess you can install this in the first step and add ansible to `$PKGS` to simplify the command line.

Dockerfile.centos7
```diff
@@ -0,0 +1,20 @@
...
+
+RUN useradd -m tester
+USER tester
+
+COPY test /test
+
+VOLUME /config /secrets /cache
+
+WORKDIR /home/tester
+CMD ["scl", "enable", "rh-git218", "/test/run-tests"]
```
tyll, line 20: The Fedora container spec uses dumb-init to ensure that the container can properly receive signals/handles zombie processes. This seems to be missing here.

README.md
```diff
@@ -21,7 +21,8 @@ change the commits in the PR.
 To make changes to the CI environment, the changes should be first committed
 to the `master` branch, and deployed to the `staging` environment.  Once they
 are confirmed to be working, the changes should be merged to the `production`
-branch and deployed to the `production` environment.
+branch and deployed to the `production` environment.  For the centos7 based tests, use `linux-system-roles-centos7` for production and
```
tyll, line 24: The previous lines seem to be wrapped but the new one is not.

README.md
```diff
@@ -150,6 +152,20 @@ a playbook and executed before the test run, or a single shell command which
 will be executed using the Ansible `raw` module (so the two examples above are
 exactly equivalent).
 
+Use the `min_ansible_version` key to indicate the minimum version of
+ansible that can be used.  For example, if you cannot use ansible 2.7
```
tyll, line 156:
```suggestion
ansible that can be used.  For example, if you need to use at least ansible 2.7
```

README.md
```diff
@@ -150,6 +152,20 @@ a playbook and executed before the test run, or a single shell command which
...
+to manage Fedora 32 hosts, use something like this:
+```json
+    {
+      "name": "fedora-32",
+      "source": "http://download.fedoraproject.org/pub/fedora/linux/releases/32/Cloud/x86_64/images/Fedora-Cloud-Base-32-1.6.x86_64.qcow2",
+      "upload_results": false,
```
tyll, line 162: This does not need to be false, does it? Also `false` should be the default, so this seems to be better:
```suggestion
      "upload_results": true,
```

README.md
```diff
@@ -440,3 +456,30 @@ linux-system-roles-12-ddtrs           1/1       Running      57         55d
...
+`run-tests` usage for the complete list.  In addition, there are some other
+env. vars which are useful for debugging:
+* `TEST_HARNESS_CONFIG_NAME` - replaces `config["name"]` - used for the name in
+the PR status message
+* `TEST_HARNESS_REPOSITORIES` - comma delimited list of repos to override
+`config["repositories"]`
+* `TEST_HARNESS_KEEP_RESULTS` - if `true`, this will cause the test not to
+delete the temp directory where it holds the results before uploading them
+
+For example, in a running OpenShift cluster, if you wanted to use a custom name for the PR status:
```
tyll, line 478: The lines seem to be wrapped inconsistently.

README.md
```diff
@@ -440,3 +456,30 @@ linux-system-roles-12-ddtrs           1/1       Running      57         55d
...
+```
+oc set env dc/linux-system-roles-staging TEST_HARNESS_CONFIG_NAME=my_custom_name
+```
+This will cause the staging pods to be redeployed and use `my_custom_name` for the status reporting.  To reset:
```
tyll, line 482: The lines seem to be wrapped inconsistently.

test/run-tests
tyll: How about changing `config_name` to `config_variant`, set it to `""` for the production instance and to `"staging"` for the staging instance and just do:

```
if config_variant:
    suffix = f" ({config_variant})"
else:
    suffix = ""
```

test/run-tests
```diff
@@ -504,7 +517,18 @@ def get_comment_commands(gh, owner, repo, pullnr):
...
+    if config_name == "linux-system-roles-test":
+        suffix = ""
+    elif config_name == "linux-system-roles-test-staging":
+        suffix = " (staging)"
+    else:
+        suffix = f" ({config_name})"
+    return f"{image_name}/{ansible_ver_str}{suffix}"
+
+
+def choose_task(gh, repos, images, config, ansible_ver_str):
```
tyll, line 531:
```suggestion
def choose_task(gh, repos, images, config, ansible_version):
```

Since this is the only ansible version in this context, there is no need to encode the type in the variable name.

test/run-tests
```diff
@@ -636,7 +662,7 @@ def make_html(source_file):
     return html_file
 
 
-def handle_task(gh, args, config, task, dry_run=False):
+def handle_task(gh, args, config, task, ansible_ver_str, dry_run=False):
```
tyll, line 665:
```suggestion
def handle_task(gh, args, config, task, ansible_version, dry_run=False):
```

test/run-tests
```diff
@@ -532,7 +556,9 @@ def choose_task(gh, repos, images, config):
             commands = get_comment_commands(gh, owner, repo, number)
 
             for image in images:
-                status_context = f"{config['name']}/{image['name']}"
+                status_context = get_status_context(
+                    config["name"], image["name"], ansible_ver_str
```
tyll, line 560:
```suggestion
                    config["name"], image["name"], ansible_version
```

test/run-tests
```diff
@@ -650,7 +676,9 @@ def handle_task(gh, args, config, task, dry_run=False):
     target_url = None
     state = None
 
-    status_context = f"{config['name']}/{task.image['name']}"
+    status_context = get_status_context(
+        config["name"], task.image["name"], ansible_ver_str
```
tyll, line 680:
```suggestion
        config["name"], task.image["name"], ansible_version
```

test/run-tests
```diff
@@ -750,7 +778,8 @@ def handle_task(gh, args, config, task, dry_run=False):
                     description += ": Error uploading results"
 
             # FIXME: workdir might be kept when python crashes
-            shutil.rmtree(workdir)
+            if os.environ.get("TEST_HARNESS_KEEP_RESULTS", "false") == "false":
```
tyll, line 781: It is probably cleaner to make this a command-line flag so all config options are in the same place.

test/run-tests
```diff
@@ -867,17 +902,63 @@ def main():
...
+    if env_repos:
+        repos = [r.split("/") for r in env_repos.split(",")]
     setup_logging(config.get("logging", {}))
 
     check_environment()
 
+    ansible_version = LooseVersion(get_ansible_version())
+    logging.info("Using Ansible version {}".format(ansible_version))
+    # this is used in PR status - context name
+    ansible_ver_str = (
```
tyll, line 933:
```suggestion
    ansible_id = (
```

This seems to be more descriptive since it does not contain the full version (so the other suggestions should probably be changed to use `ansible_id` instead of `ansible_version`.

test/run-tests
```diff
@@ -867,17 +902,63 @@ def main():
...
+        f"ansible-{ansible_version.version[0]}.{ansible_version.version[1]}"
+    )
+
+    image_patterns = args.use_images.split(",")
+    test_images = []
+    for pattern in image_patterns:
+        for image in images:
+            min_ansible_version = image.get(
```
tyll, line 941: It might be nicer to read if `LooseVersion` is applied here since the unparsed version does not seem to be needed. The logging message is also using a LooseVersion instance for the ansible version if I understand correctly.

test/run-tests
```diff
@@ -867,17 +902,63 @@ def main():
...
+                continue
+            if fnmatch.fnmatch(image["name"], pattern) or fnmatch.fnmatch(
+                image["source"], pattern
+            ):
+                if image not in test_images:
+                    test_images.append(image)
+            else:
+                logging.info(
+                    f"Skipping image {image['name']} - name and/or source "
+                    f"do not match image pattern {pattern}"
```
tyll, line 959: This seems to be more like a debug message. Also it is misleading, since if there are multiple patterns and only one matches an image, there are multiple messages that the image is skipped but it is actually used (which is not logged). To be more correct, all candidates that were not selected need to be compared at the end against all images that were selected and then a message could be logged (but this would also not log images that were skipped because of the ansible version).

test/run-tests
```diff
@@ -867,17 +902,63 @@ def main():
...
+
+    image_patterns = args.use_images.split(",")
+    test_images = []
+    for pattern in image_patterns:
+        for image in images:
+            min_ansible_version = image.get(
+                "min_ansible_version", "0"
+            )
+            if ansible_version < LooseVersion(min_ansible_version):
+                logging.info(
```
tyll, line 945: This will be logged multiple times if an image cannot be used and if there are multiple patterns. It might make sense to remove non-supported images from the pool, here. This might require changes to the for-loop definition.

Dockerfile.centos7
richm: > The Fedora container spec uses dumb-init to ensure that the container can properly receive signals/handles zombie processes. This seems to be missing here.

Not sure where to get dumb-init from for centos7.  Also, it seems to work without it - not sure what I'm missing here, or how to tell if not using it will cause problems.  Do you have a link to the Fedora container spec which talks about the use of dumb-init?

test/run-tests
richm: > How about changing `config_name` to `config_variant`, set it to `""` for the production instance and to `"staging"` for the staging instance and just do:
> 
> ```
> if config_variant:
>     suffix = f" ({config_variant})"
> else:
>     suffix = ""
> ```
> 
> This avoids any magic strings like `linux-system-roles-test` that have a special meaning.

The reason I chose `config_name` as the name of the parameter is because this exactly corresponds to `config["name"]`.  Are you suggesting that we change `config["name"]` to `config["variant"]`?

test/run-tests
richm: > Since this is the only ansible version in this context, there is no need to encode the type in the variable name.

Perhaps in this particular context, but earlier in the program execution the code has already set the precedence that the object named by `ansible_version` is indeed the numeric `LooseVersion` compatible primarily numeric (that is, looks like N.M.something where something is primarily numeric or can be alphaN or betaX), so it might be confusing for someone doing a search of this file and finding that `ansible_version` means something different depending on the context.

Perhaps you might examine the code further and find that the code in its current form has such usage where a name is used that has different meanings in different contexts, but does that mean it is important to be consistent at the expense of potential confusion?

Dockerfile.centos7
tyll: The README in https://github.com/Yelp/dumb-init explains it. Without it, there will be a lot of zombie processes in OpenShift - at least this is what we saw in the past.

test/run-tests
tyll: > The reason I chose `config_name` as the name of the parameter is because this exactly corresponds to `config["name"]`. Are you suggesting that we change `config["name"]` to `config["variant"]`?

yes, this seems to make more sense with these changes.

test/run-tests
tyll: My confusion came from that `ansible_ver_str` sounds to me like it contains a string representation of the Ansible version, so for example `"2.8.1"` but it actually contains `"ansible-2.8"`, so my new suggestion later in my comments is `ansible_id`.

test/run-tests
richm: fixed in https://github.com/linux-system-roles/test-harness/pull/87/commits/a1f7e36ec876eb927bb6ddcd25ba619b13dc88c0

test/run-tests
richm: fixed in https://github.com/linux-system-roles/test-harness/pull/87/commits/a1f7e36ec876eb927bb6ddcd25ba619b13dc88c0

test/run-tests
richm: fixed in https://github.com/linux-system-roles/test-harness/pull/87/commits/a1f7e36ec876eb927bb6ddcd25ba619b13dc88c0

test/run-tests
richm: fixed in https://github.com/linux-system-roles/test-harness/pull/87/commits/a1f7e36ec876eb927bb6ddcd25ba619b13dc88c0

test/run-tests
richm: fixed in https://github.com/linux-system-roles/test-harness/pull/87/commits/a1f7e36ec876eb927bb6ddcd25ba619b13dc88c0

test/run-tests
richm: fixed in https://github.com/linux-system-roles/test-harness/pull/87/commits/a1f7e36ec876eb927bb6ddcd25ba619b13dc88c0

test/run-tests
richm: fixed in https://github.com/linux-system-roles/test-harness/pull/87/commits/a1f7e36ec876eb927bb6ddcd25ba619b13dc88c0

test/run-tests
richm: fixed in https://github.com/linux-system-roles/test-harness/pull/87/commits/a1f7e36ec876eb927bb6ddcd25ba619b13dc88c0

test/run-tests
richm: fixed in https://github.com/linux-system-roles/test-harness/pull/87/commits/a1f7e36ec876eb927bb6ddcd25ba619b13dc88c0

README.md
richm: https://github.com/linux-system-roles/test-harness/pull/87/commits/a1f7e36ec876eb927bb6ddcd25ba619b13dc88c0

README.md
richm: https://github.com/linux-system-roles/test-harness/pull/87/commits/a1f7e36ec876eb927bb6ddcd25ba619b13dc88c0

README.md
richm: https://github.com/linux-system-roles/test-harness/pull/87/commits/a1f7e36ec876eb927bb6ddcd25ba619b13dc88c0

README.md
richm: https://github.com/linux-system-roles/test-harness/pull/87/commits/a1f7e36ec876eb927bb6ddcd25ba619b13dc88c0

README.md
richm: https://github.com/linux-system-roles/test-harness/pull/87/commits/a1f7e36ec876eb927bb6ddcd25ba619b13dc88c0

Dockerfile.centos7
richm: https://github.com/linux-system-roles/test-harness/pull/87/commits/a1f7e36ec876eb927bb6ddcd25ba619b13dc88c0

test/run-tests
richm: https://github.com/linux-system-roles/test-harness/pull/87/commits/a1f7e36ec876eb927bb6ddcd25ba619b13dc88c0

Dockerfile.centos7
richm: yep - a lot of zombie processes - doesn't make sense to me - does run-tests need to store handles to the child processes in a global array, use a SIGCHLD handler, and call `wait` on each child process in the SIGCHLD handler?  In all my years of working on openshift I have never seen this - but then none of the projects I worked on had long running container commands that did a lot of fork/exec . . .
Added back dumb-init to save debugging cycles . . .

tyll: Sorry, I don't remember the details but I guess it is related to the qcow2 inventory script forking, so the processes are not direct child processes of run-test.

test/run-tests
```diff
@@ -861,23 +893,89 @@ def main():
...
+                or fnmatch.fnmatch(image["source"], pattern)
+            ):
+                if image["name"] in test_images:
+                    logging.debug(
+                        f"Image {image['name']} will not be used for testing "
+                        f"because neither the name nor the source match the "
+                        f"image pattern {pattern}"
+                    )
+                    del test_images[image["name"]]
+                continue
```
tyll, line 973: This does not require that all patterns match for an image to be used because as soon as one image does not match, the image will be removed from the test_images, for example patterns `a,b` will remove both the images `a` and `b` IIUC.

test/run-tests
```diff
@@ -861,23 +893,89 @@ def main():
...
+    config["variant"] = os.environ.get("TEST_HARNESS_CONFIG_VARIANT", config["variant"])
+
+    env_repos = os.environ.get("TEST_HARNESS_REPOSITORIES")
+    if env_repos:
+        repos = [r.split("/") for r in env_repos.split(",")]
     setup_logging(config.get("logging", {}))
 
     check_environment()
 
+    ansible_version = LooseVersion(get_ansible_version())
```
tyll, line 940: It might make sense to move this into its own function to be able to easily create some unit tests to assert that the special cases are handled properly.

richm: ugh - this is turning into massive feature creep :-(

tyll: Another idea is to use the old logic but adjust the debug message to say "image ... does not match pattern ..." without stating whether or not it will be used for testing. Then it would be correct, too.

tyll: The idea of a unit test was to help you not to make it harder for you. :-)

richm: Sure - I'm just noting that we are getting a long way from the original intention of this PR . . .

richm: ok - the pattern match logic was fixed in https://github.com/linux-system-roles/test-harness/pull/87/commits/42f209256099726c5e8cf7b18bb3ee5fe7d59084 - I will not add unit testing in this PR - it will have to be done in another PR

tyll: seems to look correct, now

richm: squashed and rebased the PR

test/run-tests
```diff
@@ -853,31 +885,104 @@ def main():
...
+        action="store_true",
+    )
 
     args = parser.parse_args()
 
     # default values for config
     config = {"repositories": [], "images": [], "name": "linux-system-roles-test"}
 
+    # check for settings via env. vars
+    args.secrets = os.environ.get("TEST_HARNESS_SECRETS", args.secrets)
```
tyll, line 914: Sorry, did not spot this before now. It seems that config variables have the highest precendence, here. For me it seems to be more commont to have the following precendence from lowest to highest:

1) internal defaults
2) environment variables
3) config file
4) command-line arguments

not 100% sure about 2 vs 3 since it might not be that common to have that many ways to specify the same settings but since the variables are least visible, they should probably have the least precedence for user-supplied settings.

richm: I'm not sure what you mean by "config variables have the highest precedence, here.".  Are "config variables" the settings from the config.json file?  If so, there aren't any command-line arguments which allow setting those.  The only way to override those are with the `TEST_HARNESS_CONFIG_VARIANT` and `TEST_HARNESS_REPOSITORIES` env. vars.

The way it is coded now is that environment variables have the highest precedence.  For example:
```python
    args.secrets = os.environ.get("TEST_HARNESS_SECRETS", args.secrets)
```
This means - the value of `args.secrets` will be taken from the environment variable `TEST_HARNESS_SECRETS` if it is set, otherwise, use the value from the `--secrets` command-line argument if set, otherwise, use the default value.

So the precedence is:
* environment variables - highest - will override everything
* command-line arguments - can be overridden by environment variables
* config file - can be overridden by command-line arguments and environment variables
* internal defaults

This is better for the use case of running in a pod in kubernetes/openshift - it is much more difficult to change the command line arguments than to change the environment variables in a deployment.  It is a common idiom in kubernetes/openshift to allow customization via environment variables - setting an environment variable in a deployment will trigger a redeploy.

tyll: Sorry, I meant "environment variables" when I wrote "config variables" which is surprising to me, I guess code like this would be less surprising:

```python
parser.add_argument("--secrets", default=os.environ.get("TEST_HARNESS_SECRETS", "/secrets"), help="Directory with secrets")
```

> This is better for the use case of running in a pod in kubernetes/openshift - it is much more difficult to change the command line arguments than to change the environment variables in a deployment. It is a common idiom in kubernetes/openshift to allow customization via environment variables - setting an environment variable in a deployment will trigger a redeploy.

I see. But this does not require the environment variables to take precedence over command-line arguments since I guess it would be typical to deploy it without any arguments and configure it completely with variables. As far as I understand, Ansible is also allowing to use `--inventory` to override `ANSIBLE_INVENTORY` which than override the config file. So I guess I would actually find this least surprising:

* command-line arguments - will override everything:
* environment variables - highest - will override config file and defaults
* config file - can be overridden by command-line arguments and environment variables
* internal defaults

There is also a python pattern to first parse only the command-line flag to find the config file and then use this to get the defaults for a second parser run to check for remaining arguments: https://stackoverflow.com/questions/3609852/which-is-the-best-way-to-allow-configuration-options-be-overridden-at-the-comman

richm: > There is also a python pattern to first parse only the command-line flag to find the config file and then use this to get the defaults for a second parser run to check for remaining arguments

I didn't implement that, but I think this https://github.com/linux-system-roles/test-harness/pull/87/commits/fc418af4d8a50db913a4ab98b2b8f0500af98a9e makes the precedence as you suggested, and makes the usage of command line arguments and environment variables more consistent.

test/run-tests
```diff
@@ -827,39 +850,80 @@ def setup_logging(config_logging):
...
+        default=os.environ.get("TEST_HARNESS_VARIANT", ""),
+        help="Use this variant for PR test status e.g. staging",
+    )
+    parser.add_argument(
+        "--repositories",
+        default=os.environ.get("TEST_HARNESS_REPOSITORIES", None),
+        help="Comma delimited list of repositories to override config",
     )
 
     args = parser.parse_args()
```
tyll, line 929: A few lines later there is a leftover from using "name" in the config previously, that should be removed.

richm: fixed in https://github.com/linux-system-roles/test-harness/pull/87/commits/709841e06909831fcf0a8ac493bce847e5242afa

PR #91, opened 2020-07-16
paged searches for statuses

richm:
The `get_statuses` method now supports paging.  It will return all
statuses from all pages in a single return value.


nhosoi:
lgtm

PR #93, opened 2020-07-20
disable centos7/ansible 2.7 testing for now

richm:
do not deploy the centos/ansible 2.7 testing environment

LSR CI testing has a problem with hitting the github api ratelimit.
This is due to the design of the run-tests script.
Adding another deployment for ansible 2.7 testing just makes that
problem worse.  It seems that having another separate deployment
for ansible 2.7 is not the right approach.  We are in the process
of evaluating CI options.  If we stick with the current approach,
one alternative would be for the current deployment to launch
ansible-playbook for multiple versions of ansible.

However, this code may be useful, so for now not deleting it completely,
but leaving it in as disabled.

### DIFF
--- a/ansible/openshift-playbook.yml
+++ b/ansible/openshift-playbook.yml
@@ -62,6 +62,7 @@
       __test_harness_ci_obj_file: ../openshift-objects-staging.yml
       __test_harness_dc: linux-system-roles-staging
       __test_harness_bc: linux-system-roles-staging
+      __test_harness_env_is_enabled: true
     when: test_harness_use_staging | d(false) | bool
 
   - name: Ensure production is set up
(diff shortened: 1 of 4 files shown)


richm:
@pcahyna @nhosoi

nhosoi:
lgtm

pcahyna:
Can you please amend the commit message and/or the configuration YAML code to explain why are you doing the change? What was the problem with centos7/ansible 2.7 testing that led you to disabling it? It will help when someone in the future considers re-enabling it (or doing something similar).

richm:
Instead of disabling ansible 2.7 testing, we should enable ansible 2.8 testing using the new staging account

PR #94, opened 2020-07-27
run-test: add support for citool/OpenStack

ukulekek:

PR #95, opened 2020-07-29
support citool based deployment; parameterized oc objects, dockerfile

richm:
This adds support for being able to do a citool based deployment. The
openshift objects and the dockerfile has been parameterized to do this,
which should also allow us to support many more types of deployments in the
future.

There are several new ansible variables added:
`test_harness_dry_run` - default: false - if true, running the ansible 
playbook will not actually change anything in openshift, it will only dump
files which can be used with `oc apply -f`
`test_harness_dockerfile_from` - default: "docker.io/library/fedora:32" - 
User can specify the `FROM` base image
`test_harness_build_config_files` - default: [] - user can specify extra
files to be available at image build time - this is a list of strings - the
strings are the paths of the files relative to
`test_harness_config_dir`
`test_harness_dockerfile_pre_tester` - default: [] - user can specify 
additional Dockerfile commands to be executed just before switching to the
tester user e.g. `COPY` the build config files, `RUN` chown and chmod them
`test_harness_dockerfile_post` - default: [] - user can specify additional
Dockerfile commands to be executed just before the `CMD` statement e.g.
some sort of setup script
`test_harness_deploy_env` - default: {} - additional environment variables
to add to the deployment

This also use jinja2 templating to generate the openshift objects file. 
This allows us to get rid of the fixdc step.


richm:
Yes, it's sort of ugly - it depends on https://github.com/linux-system-roles/test-harness/pull/93 which needs to be reviewed and merged - imo this is a stop gap to allow us to use the existing framework to test `citool` until we can move to the new framework.  This also depends on https://github.com/linux-system-roles/test-harness/pull/94.

Q: Why did I add dry_run?
A: I needed a way to generate a citool staging deployment in files that I could edit further before using `oc apply`.  I also did not want to touch the existing deployments at all.  I can use dry_run to generate the files, edit them, and deploy them.

Q: Why not allow the user to define the output directory where the dry_run files are created?
A: This is one of those polish things.  Sure, if we decide it is worth investing even more time into the existing framework.

Q: Why so much odd code related to the Dockerfile?
A: I did not want to have two separate Dockerfiles to have to maintain separately if they only had minor differences.  I wanted the Dockerfile to be usable directly, as opposed to making the Dockerfile a jinja2 template file that would have to be processed first.  I need to amend the Dockerfile to add some internal items to it, such as the internal only base image and some other internal only commands.

Q: Why not use a yaml style inventory file so you can define lists of dicts of parameters for different environments?  Then you could use a loop to generate all of the deployments.
A: This is one of those polish things.  Sure, if we decide it is worth investing even more time into the existing framework.

Q: It sure seems hacky with all of the `*_citool` parameters . . .
A: This is one of those polish things.  Sure, if we decide it is worth investing even more time into the existing framework.  The primary concern is being able to test with citool as soon as possible.

richm:
Will need to rebase this PR since https://github.com/linux-system-roles/test-harness/pull/95/commits/6fb735e6b480a5c7e07e1d7ad5b3f8d4d4a85f78 has been dropped

richm:
Please review - this PR is required in order to move to the next step of using a new CI system

ansible/openshift-playbook.yml
```diff
@@ -73,42 +45,108 @@
       include_tasks: tasks/setup-ci-environ.yml
       vars:
         __test_harness_environ: staging
+        __test_harness_imgstream_name: linux-system-roles
         __test_harness_cm_name: config-staging
         __test_harness_cm_file: config-staging.json
-        __test_harness_ci_obj_file: ../openshift-objects-staging.yml
         __test_harness_dc: linux-system-roles-staging
         __test_harness_bc: linux-system-roles-staging
+        __test_harness_env_is_enabled: true
```
nhosoi, line 53: How is this parameter used? (I could not find a place where this param was evaluated in this pr. Same for __test_harness_ghtoken_file.)

ansible/openshift-playbook.yml
```diff
@@ -11,6 +11,12 @@
     test_harness_run_as_root: true
     test_harness_node_selector:
       system-roles-ci: "true"
+    test_harness_dockerfile_from: "docker.io/library/fedora:32"
+    test_harness_build_config_files: []
```
nhosoi, line 15: Is it ok to be empty? Or the variable will be overwritten by the inventory or command line separately?

ansible/tasks/setup-ci-environ.yml
```diff
@@ -1,32 +1,34 @@
...
-      # NOTE: This is not idempotent - it will report changed every time
-      # seems to be a problem in the k8s module
-      - name: Ensure {{ __test_harness_environ }} objects are present
-        k8s:
-          src: "{{ __test_harness_ci_obj_file }}"
-          namespace: "{{ test_harness_namespace }}"
-          state: present
+- name: create the openshift objects file from the template
+  template:
+    src: "{{ __test_harness_ci_obj_file | d(test_harness_openshift_template) }}"
```
nhosoi, line 7: There's no place to define __test_harness_ci_obj_file in this pr? Is it needed?

ansible/openshift-playbook.yml
richm: `__test_harness_env_is_enabled` - hmm - it isn't used anymore
`__test_harness_ghtoken_file` is used in `ansible/tasks/create-secrets.yml`

ansible/openshift-playbook.yml
richm: Right - and vars defined in the `vars` section here can be overridden by inventory or `-e` cmdline.

ansible/openshift-playbook.yml
nhosoi: Ahh, sorry, I missed __test_harness_ghtoken_file. Yes, it is!

richm: removed `__test_harness_env_is_enabled` in https://github.com/linux-system-roles/test-harness/pull/95/commits/5238051d32d4de4232b63fe441a6ab2681a747f7

ansible/tasks/setup-ci-environ.yml
richm: removed `__test_harness_ci_obj_file` in https://github.com/linux-system-roles/test-harness/pull/95/commits/5238051d32d4de4232b63fe441a6ab2681a747f7

PR #100, opened 2020-08-11
config and secrets dirs should be playbook vars

richm:
test_harness_secrets_dir and test_harness_config_dir should be set
via the playbook vars, not via set_fact in the flight check tasks.


richm:
note - I originally had this as part of my citool-deployment PR - need this in order to make deployments parameterizable with dry-run mode - but I thought it would be useful anyway, and having this as a separate PR makes the eventual citool-deployment PR smaller and easier to review

richm:
please review

PR #99, opened 2020-08-11
use block/always to ensure tempfile removal

richm:
put the tempfile use inside a block so that we can use always
to ensure the tempfile is removed

PR #101, opened 2020-08-11
ability to do a dry run with test_harness_dry_run=true

richm:
ability to do a dry run with test_harness_dry_run=true
depends on https://github.com/linux-system-roles/test-harness/pull/99


richm:
note - I originally had this as part of my citool-deployment PR, as it helps testing deployments of citool-staging - but I thought it would be useful anyway, and having this as a separate PR makes the eventual citool-deployment PR smaller and easier to review

richm:
please review

ansible/openshift-playbook.yml
```diff
@@ -14,9 +15,11 @@
 
   - name: Openshift pre-flight checks
     include_tasks: tasks/pre-flight-checks-openshift.yml
+    when: not test_harness_dry_run
```
tyll, line 18: How about using a block with when instead of repeating it all the time? Then it would be less likely that someone forgets to add this to new tasks in the future.

ansible/tasks/create-configmap.yml
```diff
@@ -19,8 +19,16 @@
...
+    - name: copy configmap file to {{ __test_harness_cm_name }}.yml
+      copy:
+        src: "{{ config_tempfile.path }}"
+        dest: "{{ __test_harness_cm_name }}.yml"
+      when: test_harness_dry_run | bool
+      changed_when: false
+
     - name: Remove config file
       file:
         state: absent
```
tyll, line 34: The secrets file handling looks a little bit strange, now. Where does the tempfile come from and why is it useful? Should the non-temp file be cleaned up in a non-dry run?

richm: > The secrets file handling looks a little bit strange, now.

Do you mean "configmap file"?  I'm assuming so in my following answers.

> Where does the tempfile come from

I'm not sure I understand the question.   I don't think you mean "How does the tempfile module work?"  since that's answered in the Ansible documentation.  I'm going to answer the question "Where do the contents of the tempfile come from?"

The contents are generated by the https://github.com/linux-system-roles/test-harness/blob/0c4f6fcae1183e61cf93410b9aa0f9f3ecd556d8/ansible/tasks/create-configmap.yml#L10 using the `oc create configmap --dry-run` command with the config file we want to use in the configmap.

> and why is it useful?

It is useful so that we can feed this file into the `k8s` module which will apply the changes idempotently.  It is also useful so that we can get a copy of the file for debugging and development (e.g. we want to try out some different config options in the staging environment).

> Should the non-temp file be cleaned up in a non-dry run?

I don't think it is a high priority to do this in the Ansible playbook when the user can just rm the file after use.

If someone else feels differently, please submit a followup PR.

tyll: Let me rephrase, why is a temporary file more useful than always using `"{{ __test_harness_cm_name }}.yml"` instead of copying from the temporary file to that path only in the dry-run case.

richm: > Let me rephrase, why is a temporary file more useful than always using `"{{ __test_harness_cm_name }}.yml"` instead of copying from the temporary file to that path only in the dry-run case.

So - don't use a tempfile - always write to `"{{ __test_harness_cm_name }}.yml"` - if not a dry-run, remove the file.

Sure, that would work.

richm:
and here is a diff -b showing the diffs without the whitespace only diffs:
```
diff --git a/ansible/openshift-playbook.yml b/ansible/openshift-playbook.yml
index a9d7e6e..66de881 100644
--- a/ansible/openshift-playbook.yml
+++ b/ansible/openshift-playbook.yml
@@ -2,6 +2,7 @@
 - name: Configure OpenShift deployment for linux-system-roles CI testing
   hosts: localhost
   vars:
+    test_harness_dry_run: false
     test_harness_secrets_dir: "{{ '$HOME/rhel-system-roles/test-harness-config/secrets' | expandvars }}"
     test_harness_config_dir: "{{ '$HOME/rhel-system-roles/test-harness-config/config' | expandvars }}"
     test_harness_namespace: lsr-test-harness
@@ -11,7 +12,9 @@
     test_harness_node_selector:
       system-roles-ci: "true"
   tasks:
-
+    - name: Block for tasks that only run when not using test_harness_dry_run
+      when: not test_harness_dry_run
+      block:
         - name: Openshift pre-flight checks
           include_tasks: tasks/pre-flight-checks-openshift.yml
 
diff --git a/ansible/tasks/create-configmap.yml b/ansible/tasks/create-configmap.yml
index 209f5a8..aadb14c 100644
--- a/ansible/tasks/create-configmap.yml
+++ b/ansible/tasks/create-configmap.yml
@@ -1,28 +1,21 @@
 ---
-- block:
-    - name: Create temp file for config
-      tempfile:
-        state: file
-        suffix: lsr.yml
-      register: config_tempfile
-      changed_when: false
-
-    - name: Generate config file
+- name: Generate config file
   shell: oc -n {{ test_harness_namespace }} create configmap {{ __test_harness_cm_name }} \
     --from-file=config.json={{ test_harness_config_dir }}/{{ __test_harness_cm_file }} --dry-run=true \
-        --output=yaml > {{ config_tempfile.path }}
+    --output=yaml > {{ __test_harness_cm_name }}.yml
   changed_when: false
 
-    - name: Ensure configmap is present
+- name: Ensure configmap is present
   k8s:
-        src: "{{ config_tempfile.path }}"
+    src: "{{ __test_harness_cm_name }}.yml"
     namespace: "{{ test_harness_namespace }}"
     state: present
     apply: true
+  when: not test_harness_dry_run
 
-  always:
-    - name: Remove config file
+- name: Remove config file
   file:
     state: absent
-        path: "{{ config_tempfile.path }}"
+    path: "{{ __test_harness_cm_name }}.yml"
   changed_when: false
+  when: not test_harness_dry_run
diff --git a/ansible/tasks/create-secrets.yml b/ansible/tasks/create-secrets.yml
index 64c33b2..0c91dc4 100644
--- a/ansible/tasks/create-secrets.yml
+++ b/ansible/tasks/create-secrets.yml
@@ -1,13 +1,5 @@
 ---
-- block:
-    - name: Create temp file for secrets
-      tempfile:
-        state: file
-        suffix: lsr.yml
-      register: secrets_tempfile
-      changed_when: false
-
-    - name: Generate secrets file
+- name: Generate secrets file
   shell: oc -n {{ test_harness_namespace }} create secret generic {{ __test_harness_secret_name }} \
     --from-file=github-token={{ test_harness_secrets_dir }}/{{ __test_harness_ghtoken_file }} \
     --from-file=id_rsa={{ test_harness_secrets_dir }}/id_rsa \
@@ -15,18 +7,19 @@
     --from-file=openstack-psi={{ test_harness_secrets_dir }}/openstack-psi \
     --from-file=system-roles-os={{ test_harness_secrets_dir }}/system-roles-os \
     --dry-run=true \
-        --output=yaml > {{ secrets_tempfile.path }}
+    --output=yaml > {{ __test_harness_secret_name }}.yml
   changed_when: false
 
-    - name: Ensure secrets are present
+- name: Ensure secrets are present
   k8s:
-        src: "{{ secrets_tempfile.path }}"
+    src: "{{ __test_harness_secret_name }}.yml"
     namespace: "{{ test_harness_namespace }}"
     state: present
+  when: not test_harness_dry_run
 
-  always:
-    - name: Remove secrets file
+- name: Remove secrets file
   file:
     state: absent
-        path: "{{ secrets_tempfile.path }}"
+    path: "{{ __test_harness_secret_name }}.yml"
   changed_when: false
+  when: not test_harness_dry_run
diff --git a/ansible/tasks/setup-ci-environ.yml b/ansible/tasks/setup-ci-environ.yml
index 986a457..01ce7ff 100644
--- a/ansible/tasks/setup-ci-environ.yml
+++ b/ansible/tasks/setup-ci-environ.yml
@@ -2,6 +2,9 @@
   - name: Ensure {{ __test_harness_environ }} configmap is present
     include_tasks: create-configmap.yml
 
+  - name: Block for tasks that only run when not using test_harness_dry_run
+    when: not test_harness_dry_run
+    block:
       # NOTE: This is not idempotent - it will report changed every time
       # seems to be a problem in the k8s module
       - name: Ensure {{ __test_harness_environ }} objects are present
```

richm:
ok to merge?

PR #105, opened 2020-09-02
increase num of statuses to 99 - 60 is too small

richm:
Seeing error messages like this in production:
```
ERROR: Total number of statuses 61 exceeds max_num_statuses 60 statuses for xxx - please increase max_num_statuses
```
I have temporarily fixed production by using the env var setting, but
this will be wiped out at the next redeploy.

### DIFF
--- a/test/run-tests
+++ b/test/run-tests
@@ -1040,8 +1040,8 @@ def main():
     parser.add_argument(
         "--max-num-statuses",
         type=int,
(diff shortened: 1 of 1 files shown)


richm:
> lgtm
> Just out of curiosity, why 99, instead of, e.g., 999? :)

github imposes a hard limit of 100

PR #107, opened 2020-10-12
Use deploy_replicas* also for centos7 deployments

pcahyna:
Synchronizes centos7 deployments with the standard deployments. But keep the default for centos7 production as 1 to avoid a thundering herd of production pods when the variable is unset (default for standard production is 8).

### DIFF
--- a/ansible/openshift-playbook.yml
+++ b/ansible/openshift-playbook.yml
@@ -115,7 +115,7 @@
         __test_harness_bc: linux-system-roles-centos7-staging
         __test_harness_secret_name: secrets-staging
         __test_harness_ghtoken_file: github-token-staging
-        __test_harness_deploy_replicas: 1
+        __test_harness_deploy_replicas: "{{ test_harness_deploy_replicas_staging | d(1) }}"
         __test_harness_dockerfile: ../Dockerfile.centos7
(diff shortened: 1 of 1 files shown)


ansible/openshift-playbook.yml
```diff
@@ -115,7 +115,7 @@
         __test_harness_bc: linux-system-roles-centos7-staging
         __test_harness_secret_name: secrets-staging
         __test_harness_ghtoken_file: github-token-staging
-        __test_harness_deploy_replicas: 1
+        __test_harness_deploy_replicas: "{{ test_harness_deploy_replicas_staging | d(1) }}"
```
jharuda, line 118: What about 4 variables for replicas?? test_harness_deploy_replicas_staging, test_harness_deploy_replicas,  **test_harness_deploy_replicas_centos_staging, test_harness_deploy_centos_replicas** to make it more configurable? 

@richm

richm (approved):
lgtm

ansible/openshift-playbook.yml
pcahyna: I am going to merge the PR as it is. The change you suggest can be done in a future PR.

PR #108, opened 2020-10-21
openshift-playbook: Add staging env variable

jharuda:


ansible/openshift-playbook.yml
```diff
@@ -58,7 +59,7 @@
         __test_harness_dockerfile_pre_tester: "{{ test_harness_dockerfile_pre_tester }}"
         __test_harness_build_config_files: "{{ test_harness_build_config_files }}"
         __test_harness_dockerfile_post: "{{ test_harness_dockerfile_post }}"
-        __test_harness_deploy_env: "{{ test_harness_deploy_env }}"
+        __test_harness_deploy_env_staging: "{{ test_harness_deploy_env_staging }}"
```
pcahyna, line 62: This won't work, because `__test_harness_deploy_env_staging` isn't used anywhere.

PR #110, opened 2020-10-21
Derive image URL from compose URL

### LINKED ISSUE #15: Add support to test against updated releases
AFAICS the test harness only checks against the base releases but does not update the images.

pcahyna:
Can be used to support images whose URL is changing dynamically, like latest Fedora cloud images.

Introduces a new option "compose" in the image dictionary, which can be used instead of "source" and contains the compose URL.

Optionally it can be qualified by "variant" when the compose contains images of multiple variants and specifying only compose would be ambiguous.

Also properly report error when image can't be downloaded as an error not test failure.

Closes #15

### DIFF
--- a/test/run-tests
+++ b/test/run-tests
@@ -33,6 +33,7 @@ import yaml
 # use LooseVersion to handle alphas, betas, other pre-release versions
 from distutils.version import LooseVersion
 from distutils.util import strtobool
+from compose2image import composeurl2images
 
 HOSTNAME = socket.gethostname()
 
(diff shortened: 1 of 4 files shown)


test/run-tests
```diff
@@ -328,6 +329,25 @@ class Task:
...
+            if len(image_urls) == 1:
+                return image_urls[0]
+            else:
+                if image_urls:
+                    logging.error(
+                        f"ERROR: Multiple images found: {image_urls}"
+                        "in compose {compose_url}"
+                    )
+                else:
+                    logging.error(f"ERROR: no image found in compose {compose_url}")
```
richm, line 349: do we need to check in case there is no `source` and there is no `compose`?

richm: or is that handled below with the
```python
            if not image_url:
                return None
```
?

pcahyna: That was indeed the idea. But a more explicit error message in the log would not hurt.

test/run-tests
```diff
@@ -1140,7 +1163,7 @@ def main():
         pattern_matched = False
         for pattern in image_patterns:
             if fnmatch.fnmatch(image["name"], pattern) or fnmatch.fnmatch(
-                image["source"], pattern
+                image.get("source", ""), pattern
```
richm, line 1166: should we try to match on source or compose here?  I suppose the current code doesn't handle openstack images anyway

pcahyna: Do we use it to match on anything else than name in practice?

test/run-tests
pcahyna: fixed by rebase

test/run-tests
richm: Probably not - I would say this is ok - and if a dev needs to do something else with pattern (e.g. for debugging) that dev can enhance the code to do the needful

richm (approved):
lgtm

jharuda:
LGTM

jharuda (approved):

PR #111, opened 2020-10-22
run-tests: Skip unsupported distributions

jharuda:


test/run-tests
```diff
@@ -657,6 +658,26 @@ def get_status_context(variant, image_name, ansible_id):
...
 
+def is_role_distro_supported(image, repo_dir):
+    """
+    Return True if image is supported. Otherwise False.
+    """
+    distro = image["name"].split("-")[0]
+    version = image["name"].split("-")[1]
+    if version == 'x':
+        image_filename=(
+            image['source'].rsplit('/', 1)[1]
```
richm, line 669: How would this work with https://github.com/linux-system-roles/test-harness/commit/95372553cc0dad26ac045cd2a3583015c0d73510 if there isn't a `image['source']` field?  I suppose for now it is ok since the only `-x` version we have is with rhel-x - but please consider how this would work if there is `image['compose']` and no `image['source']`

test/run-tests
```diff
@@ -684,24 +705,34 @@ def choose_task(gh, repos, images, config, ansible_id, args):
...
-                )
-                status = statuses.get(status_context)
+            # We need PR repository to check meta in is_role_distro_support()
+            with checkout_repository(
+                owner, repo, f"pull/{pull}/head"
+            ) as repo_tmp_dir:
+                for image in images:
+                    if is_role_distro_supported(image, repo_tmp_dir):
+                        logging.debug(
+                            f"Unsupported Role. Skipping role {repo} with image {image['source']}."
```
richm, line 715: I suggest a message like "The role {repo} does not support platform {image['name']}.  Skipping."

test/run-tests
jharuda: Good point. I can extract the `get_url()` function from the Task class or simply get the first numbers in compose URL. The second option is not so unique.

pcahyna: Can't you just call `get_url()`? It would require to move the check after Task is created, would it be a problem?

jharuda: Fixed. I use `get_url()`. The function `is_role_distro_supported()` is a part of the Task class.

jharuda:
Here are some tests in the ci-testing repo:

https://github.com/linux-system-roles/ci-testing/pull/4
https://github.com/linux-system-roles/ci-testing/pull/5
https://github.com/linux-system-roles/ci-testing/pull/6

jharuda:
Please review the current commit #7475706.

test/run-tests
```diff
@@ -353,6 +354,42 @@ class Task:
...
             )
 
+    def is_role_distro_supported(self):
+        """
+        Return True if image is supported. Otherwise False.
+        """
+        distro = self.image["name"].split("-")[0]
+        version = self.image["name"].split("-")[1]
+        if version == "x":
+            image_filename = self.get_url().rsplit("/", 1)[1]
```
pcahyna, line 364: Please use the same method as https://github.com/linux-system-roles/test-harness/blob/3796e3f92d2c71fc74e2d1889edc2ad82966e415/test/run-tests#L257

pcahyna (approved):

test/run-tests
```diff
@@ -353,6 +354,44 @@ class Task:
                 f"in image {self.image['name']}"
             )
 
+    def is_role_distro_supported(self):
+        """
+        Return True if image is supported. Otherwise False.
+        """
+        distro = self.image["name"].split("-")[0]
+        version = self.image["name"].split("-")[1]
```
richm, line 362:
```python
distro, version = self.image["name"].split("-")
```

richm: This also has the added benefit that it will throw an exception if the format of name is not in the expected `DISTRO-VERSION` format

richm:
one minor comment - otherwise lgtm

test/run-tests
jharuda: OK. Fixed in #6dc7afd.

richm (approved):
lgtm

richm:
You'll need to rebase on top of https://github.com/linux-system-roles/test-harness/pull/112

PR #113, opened 2020-10-29
RHELPLAN-42929 - Collections - script - add CI testing for collection

nhosoi:
If `test_collections` is set to true in the config file (`config.json`) - [0], `lsr_role2collection.py` is downloaded and the role of being tested is converted to the collection format. The role's tests directory is placed outside of the collection and the tests playbooks are executed from there.

[0] default to false.

This PR depends upon these 2 PRs to fix lsr_role2collection.py.
https://github.com/linux-system-roles/auto-maintenance/pull/27
https://github.com/linux-system-roles/auto-maintenance/pull/28

QUESTION: I have tested only with KVM_BACKEND. Could someone tell me how to test locally with CITOOL_BACKEND?

### DIFF
--- a/test/run-tests
+++ b/test/run-tests
@@ -392,7 +392,13 @@ class Task:
             return None
 
     def run(
-        self, artifactsdir, private_artifactsdir, cachedir, backend, inventory=None
+        self,
+        artifactsdir,
+        private_artifactsdir,
+        cachedir,
+        backend,
+        inventory=None,
+        test_collections=False,
     ):
         """
         Runs the task and puts results into @artifactsdir. Puts non-public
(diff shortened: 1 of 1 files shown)


test/run-tests
```diff
@@ -467,7 +497,15 @@ class Task:
             # used to be tested by running all playbooks `test/test_*.yml`.
             # Support both, but prefer the standard way. Can be removed once
             # all repos are moved over.
-            playbookglob = f"{sourcedir}/tests/tests*.yml"
+            if test_collections:
+                playbookglob = (
+                    os.environ.get("COLLECTION_TESTS_DEST_PATH")
```
richm, line 502: use `os.environ["COLLECTION_TESTS_DEST_PATH"]` here to catch bugs - if it is not set it will raise an error instead of silently use `/tests/..` as the path.

test/run-tests
```diff
@@ -494,6 +532,13 @@ class Task:
                     # invocation of ansible-playbook, so that that it applies
                     # to the same VM as the test playbook.
                     if backend == CITOOL_BACKEND:
+                        testenv = {}
+                        if test_collections:
+                            testenv = {
+                                "ANSIBLE_COLLECTIONS_PATHS": collection_paths,
+                            }
+                        else:
+                            testenv = {}
```
richm, line 541: You don't need the `else` clause here

test/run-tests
```diff
@@ -521,20 +566,29 @@ class Task:
                             "test-schedule-runner-sti",
                             "test-schedule-runner",
                             "test-schedule-report",
+                            env=testenv,
                             check=False,
                             cwd=os.path.dirname(playbook),
                         )
                     elif backend == KVM_BACKEND:
```
richm, line 573:
```python
    testenv = {
        "TEST_SUBJECTS": image_path,
        "TEST_ARTIFACTS": artifactsdir,
     }
     if test_collections:
         testenv["ANSIBLE_COLLECTIONS_PATHS"] = collection_paths
```

richm:
lgtm - squash and merge

nhosoi:
Thank you for reviewing this PR, @richm! It looks I do not have a privilege to merge. May I ask you to do so?
Thanks!

PR #114, opened 2020-11-04
Keep only latest image file in the cache

pcahyna:
Image URLs are changing now, and the image file has been kept in the cache under a name derived from the URL. This has led to obsolete images images piling up in the cache as the URL has kept changing.

Fix by saving the image file under a filename derived from the image label (like 'fedora-32'), not from the URL. This way there can be only one image in the cache for the given URL.

To determine if the cached file is up to date record its URL in an extended attribute. If the recorded URL does not match the current URL, the image is obsolete and will be redownloaded.

Fixes a problem introduced in #110.

### DIFF
--- a/test/run-tests
+++ b/test/run-tests
@@ -9,6 +9,7 @@ import html
 import json
 import logging
 import os
+import errno
 import random
 import re
 import signal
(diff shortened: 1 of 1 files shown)


richm (approved):
lgtm

jharuda (approved):
LGTM

PR #117, opened 2020-11-06
runtest: Fix "No module named 'lsr_role2collection" error

### LINKED ISSUE #116: Exception when running tests: No module named 'lsr_role2collection'
We need to store the file `lsr_role2collection.py` in correct location to be able to use from `lsr_role2collection import role2collection`. The current code stores it in home directory, but `runtest` can be elsewhere. It cause "Exception when running tests: No module named 'lsr_role2collection'"

https://github.com/linux-system-roles/test-harness/blob/master/test/run-tests#L454

jharuda:
resolves #116

### DIFF
--- a/test/run-tests
+++ b/test/run-tests
@@ -451,7 +451,11 @@ class Task:
             if test_collections:
                 r2c_url = "https://raw.githubusercontent.com/linux-system-roles/auto-maintenance/master/lsr_role2collection.py"  # noqa:E501
                 r2c = requests.get(r2c_url, allow_redirects=True)
-                open("lsr_role2collection.py", "wb").write(r2c.content)
+                role2collection_path = (
+                    f"{ os.path.dirname(os.path.realpath(__file__)) }"
+                    "/lsr_role2collection.py"
+                )
+                open(role2collection_path, "wb").write(r2c.content)
                 from lsr_role2collection import role2collection
 
                 os.environ["COLLECTION_SRC_PATH"] = sourcedir


nhosoi:
That's exactly what I was going to add. Thanks, @jharuda!
Before merging could you please the black error?
Other than that, lgtm.

nhosoi (approved):
lgtm

nhosoi (approved):

nhosoi:
@jharuda, thanks for the fix. But when I tested this code in the container, it fails to write the role2collection script in the /test dir... (;_;)
```
1effc349b694: linux-system-roles/logging: pull #186 (405f09a) on centos-8
======================================================================
+ git init --quiet /tmp/tmp64rjo4jn
+ GIT_TERMINAL_PROMPT=0 git -C /tmp/tmp64rjo4jn fetch --quiet --depth=1 https://github.com/linux-system-roles/logging pull/186/head
+ git -C /tmp/tmp64rjo4jn checkout --quiet FETCH_HEAD
Traceback (most recent call last):
  File "/test/run-tests", line 999, in handle_task
    test_collections=config.get("test_collections", False),
  File "/test/run-tests", line 458, in run
    open(role2collection_path, "wb").write(r2c.content)
PermissionError: [Errno 13] Permission denied: '/test/lsr_role2collection.py'
```
To workaround this failure, we could do either of these, I'd think.
1) checking in lsr_role2collection.py in test-harness/test,
2) allowing run-tests to write in /test in the container.
What would be better? Or could there be even a better solution? Thanks.

richm:
> checking in lsr_role2collection.py in test-harness/test,

I would rather not.  We would have to move it there from auto-maintenance (then have to move it somewhere else if we ever completely move to baseosci) to avoid having two copies.

> allowing run-tests to write in /test in the container

Possible - would need to change the deployment config volume permissions.

> putting run-tests in some writable & imporable location

Yes.  Then you can set `PYTHONPATH=/path/to/location`

PR #120, opened 2020-11-13
runtest: Skip rhel-x in citool backend

jharuda:


richm (approved):
lgtm

pcahyna:
Sorry for being late to the party, but do we need such special case in the code? Can't we skip anything that does not have `openstack_image` defined and then undefine it on any image where we don't want citool runs?

PR #123, opened 2020-11-16
Do not pull images from Docker.io 

jharuda:
docker.io recently changed their free tier download rates. Find an alternative provider.

https://www.docker.com/increase-rate-limits

### DIFF
--- a/ansible/openshift-playbook.yml
+++ b/ansible/openshift-playbook.yml
@@ -11,7 +11,7 @@
     test_harness_run_as_root: true
     test_harness_node_selector:
       system-roles-ci: "true"
-    test_harness_dockerfile_from: "docker.io/library/fedora:32"
+    test_harness_dockerfile_from: "registry.fedoraproject.org/fedora:32"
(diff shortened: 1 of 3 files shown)


ansible/openshift-playbook.yml
```diff
@@ -11,7 +11,8 @@
     test_harness_run_as_root: true
     test_harness_node_selector:
       system-roles-ci: "true"
-    test_harness_dockerfile_from: "docker.io/library/fedora:32"
+    test_harness_dockerfile_from: "registry.fedoraproject.org/fedora:32"
+    test_harness_dockerfile_from_centos: "registry.centos.org/centos/centos:7"
```
richm, line 15: this should be `test_harness_dockerfile_from_centos7` to be consistent with the naming convention used by other variables which are specific to centos7 and have a `_centos7` suffix

jharuda: Fixed in #96902af

richm (approved):
lgtm

PR #130, opened 2020-12-05
nightly composes: Don't raise an exception if we find more images

jharuda:
Use a less restrictive way to get an image from a compose URL.
The latest .qcow2 image is the last in the alphabet and first in the list of images.

### DIFF
--- a/test/compose2image.py
+++ b/test/compose2image.py
@@ -40,7 +40,9 @@ def composeurl2images(
             if len(subvariantmatch) > 0:
                 candidates = subvariantmatch
 
-    return [(composepath + qcow2[0].path) for qcow2 in candidates]
+    image_urls = [(composepath + qcow2[0].path) for qcow2 in candidates]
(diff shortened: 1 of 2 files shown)


test/compose2image.py
```diff
@@ -40,7 +40,9 @@ def composeurl2images(
             if len(subvariantmatch) > 0:
                 candidates = subvariantmatch
 
-    return [(composepath + qcow2[0].path) for qcow2 in candidates]
+    image_urls = [(composepath + qcow2[0].path) for qcow2 in candidates]
+    return sorted(image_urls, reverse=True)
```
richm, line 44: How does this guarantee that the latest image is the first one in the list?

jharuda: @richm There is no 100% guarantee. This implementation expects the new nightly compose to have the same format as the older one and to have alphabetically bigger character in the image name. So it uses only `sorted()` function.

PR #137, opened 2021-01-19
runtest: Add support for alias image URLs

jharuda:
It attempts to add functionality to get the latest image from a URL, where URL (`source` field in the config) is a alias for the latest image.
It is useful for CentOS images.
It uses the metadata of the web resource to distinguish whether a new image needs to be downloaded.

### DIFF
--- a/test/run-tests
+++ b/test/run-tests
@@ -49,6 +49,7 @@ CITOOL_COMMAND = "/entrypoint.sh"
 
 # https://www.freedesktop.org/wiki/CommonExtendedAttributes/
 URL_XATTR = "user.xdg.origin.url"
+DATE_XATTR = "user.dublincore.date"
 
 
 class TransientErrorWaitTime:
@@ -249,18 +250,30 @@ def run(*argv, env=None, check=True, cwd=None, return_stdout=False):
     return result
 
 
-def origurl(path):
-    """
-    Returns the original URL that a given file was downloaded from.
-    """
-
+def get_metadata_from_file(path, attr_key):
     try:
-        urlbytes = os.getxattr(path, URL_XATTR)
+        mdbytes = os.getxattr(path, attr_key)
     except OSError as e:
         if e.errno == errno.ENODATA:
             return None
         raise
-    return os.fsdecode(urlbytes)
+    return os.fsdecode(mdbytes)
+
+
+def image_source_last_modified_by_file_metadata(path):
+    return get_metadata_from_file(path, DATE_XATTR) if os.path.exists(path) else ""
+
+
+def origurl(path):
+    """
+    Returns the original URL that a given file was downloaded from.
+    """
+    return get_metadata_from_file(path, URL_XATTR)
+
+
+def get_metadata_from_url(url, metadata_key):
+    with urllib.request.urlopen(url) as url_response:
+        return dict(url_response.getheaders())[metadata_key]
 
 
 def fetch_image(url, cache, label):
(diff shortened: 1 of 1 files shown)


jharuda:
Ideas for what can be done differently:
- I can remove checking `origurl` != URL because we use last modified date
- I can add a new optional option in the `config.json` and call it `alias_url`. It will have the similar functionality as `source`, but in the code it will use `last_modified` attribute comparison. If there is no `alias_url` option in the config, it will use the current comparison ( (`origurl` != URL).

test/run-tests
```diff
@@ -296,6 +315,7 @@ def fetch_image(url, cache, label):
             return None
 
         os.setxattr(image_tempfile.name, URL_XATTR, os.fsencode(url))
+        os.setxattr(image_tempfile.name, DATE_XATTR, image_last_modified_by_src)
```
richm, line 318: what is the format for `image_last_modified_by_src`?  Do you have to use `os.fsencode` to encode the value before storing it in the xattr?

pcahyna:
I don't understand from the description what is it supposed to do. Can you provide an example? Also, is it a new feature or does it fix a misbehavior in some cases?

test/run-tests
jharuda: The `image_last_modified_by_src` variable stores strings like "Wed, 22 Apr 2020 12:24:13 GMT". It has `str` filetype so in the latest commit I added `os.fsencode` to convert it to bytes-like object.

The downloaded image file metadata looks like:
$ getfattr -d CentOS-7-x86_64-GenericCloud.qcow2c
# file: CentOS-7-x86_64-GenericCloud.qcow2c
user.xdg.origin.url="https://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud.qcow2c"
user.dublincore.date="Wed, 22 Apr 2020 12:24:13 GMT"

jharuda:
@pcahyna  It is a new feature for selecting CentOS images.
Currently we support 2 methods how to define/find image: 1) compose URL and 2) source URL. When we use the `source` URL method, we expect that the uploaded image will not be changed and(or) updated.
The point of this PR is to add support for third method. In the third method I extended `source` URL functionality to support alias URL (by alias URL I mean for example https://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud.qcow2c). The third method is useful because we will not need to manually update image in the config to CentOS-7-x86_64-GenericCloud-1907.qcow2c, then to CentOS-7-x86_64-GenericCloud-2003.qcow2c then to CentOS-7-x86_64-GenericCloud-2009.qcow2c ... when there is available newer CentOS image.

richm:
please fix the tox issues

jharuda:
> please fix the tox issues

done

test/run-tests
```diff
@@ -249,20 +250,32 @@ def run(*argv, env=None, check=True, cwd=None, return_stdout=False):
...
 
-def origurl(path):
-    """
-    Returns the original URL that a given file was downloaded from.
-    """
-
+def get_metadata_from_file(path, attr_key):
     try:
-        urlbytes = os.getxattr(path, URL_XATTR)
+        urlbytes = os.getxattr(path, attr_key)
```
richm, line 255: this isn't `urlbytes` anymore - I suggest `mdbytes` for metadata bytes

jharuda: Yes, `mdbytes` is better. I fixed that.

richm (approved):
lgtm

PR #141, opened 2021-02-10
runtest: Add support for image names with two hyphens (rhel-8-y)

jharuda:
We currently support image formats like: `rhel-8` and `rhel-x`.

This PR adds support for the format "rhel-8-y". The `-y` part is ignored.


richm (approved):
lgtm

PR #144, opened 2021-02-18
test-harness - Fixing a CI test failure

nhosoi:
In the collection converted tests, the tempdirs to store converted
role and CI test files were not cleaned up every test cycle (per pr),
but only at the end. It left the extra files from the previous run
and when the leftover file is a test script, it caused a test failure.

With this fix, the tempdirs are generated and cleaned up for each pr.

### DIFF
--- a/test/run-tests
+++ b/test/run-tests
@@ -447,6 +447,7 @@ class Task:
         backend,
         inventory=None,
         test_collections=False,
+        keep_results=False,
     ):
         """
         Runs the task and puts results into @artifactsdir. Puts non-public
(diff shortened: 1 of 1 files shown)


test/run-tests
```diff
@@ -550,7 +559,17 @@ class Task:
...
+            # If test_collections is true, converts the role to the collections format.
             if test_collections:
+                from importlib import import_module
+
+                _lsr = import_module("lsr_role2collection")
+                _lsr.role2collection()
+                collection_paths = (
+                    os.environ["COLLECTION_DEST_PATH"]
+                    + ":~/.ansible/collections:/usr/share/ansible/collections"
+                )
```
richm, line 571: You don't have to do this for every iteration - you only have to do this once

nhosoi: Once per pull request == task.run, isn't it?

test/run-tests
```diff
@@ -681,6 +700,10 @@ class Task:
                         return False
 
                     print("SUCCESS")
+
+            if test_collections and not keep_results:
+                cleanup_collection_tempdirs()
```
richm, line 705: This is problematic because it will remove the lsr_role2collection script - you just want to clean up the collection converted role and tests here, not the lsr script too - see #145

nhosoi:
Closing this pr in favor of https://github.com/linux-system-roles/test-harness/pull/145.

PR #145, opened 2021-02-18
make a new tempdir for collections, tests for each iteration

richm:
Do not reuse the directories used for converting the role and
tests to collection format.  Instead, for each set of test
playbooks, create new temporary directories for the role and
tests, and remove them when finished.


nhosoi (approved):
lgtm

PR #147, opened 2021-03-24
Use the same qcow2 inventory/VM provision script in centos7 as in Fedora

jharuda:
Use python3.6 version of packages

### DIFF
--- a/Dockerfile.centos7
+++ b/Dockerfile.centos7
@@ -2,12 +2,15 @@ FROM registry.centos.org/centos/centos:7
 
 RUN yum update -y && PKGS="centos-release-ansible-28 centos-release-scl-rh" && \
     yum install -y $PKGS && rpm -V $PKGS && \
-    PKGS="ansible epel-release python2-jmespath rsync python-netaddr" && yum install -y $PKGS && rpm -V $PKGS && \
-    PKGS="rh-git218 python3-pip standard-test-roles python2-fmf seabios-bin" && \
+    PKGS="ansible epel-release rsync" && yum install -y $PKGS && rpm -V $PKGS && \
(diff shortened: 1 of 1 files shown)


Dockerfile.centos7
```diff
@@ -21,4 +21,5 @@ VOLUME /config /secrets /cache
 
 WORKDIR /home/tester
 ENTRYPOINT ["/usr/local/bin/dumb-init", "--"]
+CMD ["cp", "/test/standard-inventory-qcow2-python3", "/usr/share/ansible/inventory/standard-inventory-qcow2"]
```
richm, line 24: Rather than having a copy of the script in our repo that we have to maintain, I would rather just download the script from the source when we build the container e.g. `curl -s -o /usr/share/ansible/inventory/standard-inventory-qcow2 https://pagure.io/standard-test-roles/raw/ce0ce5eb049c886a0cb4ecbd3f4c9b409693ba4a/f/inventory/standard-inventory-qcow2`
One wrinkle is that you cannot apparently download from a git tag from pagure like you can from github, so that's why the link above uses the full git commit hash `ce0ce5eb049c886a0cb4ecbd3f4c9b409693ba4a` instead of a tag.  If you go to the Releases page - https://pagure.io/standard-test-roles/releases - you can see that the latest release is `4.10`  - if you go to the files, you can see that it corresponds to the commit hash.

jharuda:
Note: We need to install the `epel-release` package before `python36-jmespath` and `python36-netaddr` because these packages are only available in EPEL. The older Dockerfile.centos7 works for `python2-jmespath` and `python-netaddr` because it probably do not install these packages from EPEL.

Dockerfile.centos7
jharuda: It make sense. I updated it the latest commit.

richm (approved):
lgtm

PR #148, opened 2021-03-27
netaddr and jmespath are for python2 for ansible

richm:
netaddr and jmespath are for python2 for ansible, not for python3
for the standard test script

### DIFF
--- a/Dockerfile.centos7
+++ b/Dockerfile.centos7
@@ -1,9 +1,9 @@
 FROM registry.centos.org/centos/centos:7
 
-RUN yum update -y && PKGS="centos-release-ansible-28 centos-release-scl-rh" && \
+RUN yum update -y && PKGS="centos-release-ansible-28 python2-jmespath python-netaddr centos-release-scl-rh" && \
     yum install -y $PKGS && rpm -V $PKGS && \
     PKGS="ansible epel-release rsync" && yum install -y $PKGS && rpm -V $PKGS && \
-    PKGS="python36-jmespath python36-netaddr python36-fmf" && yum install -y $PKGS && rpm -V $PKGS && \
+    PKGS="python36-fmf" && yum install -y $PKGS && rpm -V $PKGS && \
     PKGS="rh-git218 python3-pip standard-test-roles seabios-bin" && \
     yum -y install $PKGS && rpm -V $PKGS && \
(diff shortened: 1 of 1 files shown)


pcahyna:
Good catch, we are now using both Python versions in the container at the same time.

I have not been aware of the need for the netaddr package - I suppose it is for the [ipaddr filter](https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters_ipaddr.html)? How is it possible that roles on RHEL and Fedora do not suffer from the lack of this dependency, as they did for jmespath? (Neither the system roles RPM nor the Ansible RPM depend on netaddr.)

richm:
> Good catch, we are now using both Python versions in the container at the same time.
> 
> I have not been aware of the need for the netaddr package - I suppose it is for the [ipaddr filter](https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters_ipaddr.html)? How is it possible that roles on RHEL and Fedora do not suffer from the lack of this dependency, as they did for jmespath? (Neither the system roles RPM nor the Ansible RPM depend on netaddr.)

The vpn role uses the ipaddr filter.

richm:
> > Good catch, we are now using both Python versions in the container at the same time.

Well, we have been for a long time.  The test/run-tests script requires python3, and we run ansible 2.8 with python2.

> > I have not been aware of the need for the netaddr package - I suppose it is for the [ipaddr filter](https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters_ipaddr.html)? How is it possible that roles on RHEL and Fedora do not suffer from the lack of this dependency, as they did for jmespath? (Neither the system roles RPM nor the Ansible RPM depend on netaddr.)

So we'll have to add a netaddr dependency on the Fedora and rhel 8.5 packages.

> 
> The vpn role uses the ipaddr filter.

jharuda:
Do we use `python-jmespath` only in network role?

richm:
> Do we use `python-jmespath` only in the network role?

No - I see usage of `json_query` in storage and timesync as well.

richm:
ok to merge?  This fixes the errors I was seeing in timesync https://github.com/linux-system-roles/timesync/pull/110

jharuda (approved):
LGTM

PR #149, opened 2021-04-01
do not report failure if only error is uploading results

richm:
The github status should reflect the test status, not the result
uploading status.  Upload failures will be logged and reported
elsewhere.


nhosoi (approved):

jharuda (approved):
LGTM

pcahyna:
Looking at the code, it was not reported as "failure" but as "error", which was more appropriate. Still, it is better to report success and log the problem elsewhere.

PR #153, opened 2021-07-12
the file meta/runtime.yml is now needed for collections

richm:
due to the recent commit https://github.com/linux-system-roles/auto-maintenance/pull/83
we need to add the runtime.yml file for lsr_role2collection to use


nhosoi (approved):
lgtm

PR #161, opened 2021-12-01
subrole prefix with collection; debug http

richm:
add subrole prefix when converting to collection
add debugging for github http traffic


richm:
this goes with https://github.com/linux-system-roles/metrics/pull/99

nhosoi (approved):
lgtm

Thanks for the fix!

PR #169, opened 2022-04-29
Remove snapshot, yum caches by default

richm:
Since we are tight on disk space for /cache, remove the .snap and
yum cache files by default after the PR test run.
Added parameter `--keep-snap-cache` in case the user wants
to keep the cache files.

### DIFF
--- a/test/run-tests
+++ b/test/run-tests
@@ -750,6 +750,16 @@ class Task:
                 logging.debug(f"test passed for {artifactsdir}")
             else:
                 logging.debug(f"test failed for {artifactsdir}")
+            if not args.keep_snap_cache:
+                # We are currently tight on disk space in /cache
+                # remove the snap and cache files after the test by default.
+                # Revisit this if we change the implementation.
+                files_to_rm = glob.glob(os.path.join(args.cache, "*.snap"))
+                files_to_rm.extend(glob.glob(os.path.join(args.cache, "*_yum_cache")))
+                files_to_rm.extend(glob.glob(os.path.join(args.cache, "*_yum_varlib")))
+                for frm in files_to_rm:
+                    logging.debug(f"removing {frm}")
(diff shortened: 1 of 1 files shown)


nhosoi (approved):
lgtm

Actually, I was going to ask a question related to this :).
Why do we need to have the same looking files?
```
-rw-r--r--. 1 nhosoi nhosoi  169672704 Apr 29 12:27 centos-8.qcow2.snap
-rw-r--r--. 1 nhosoi nhosoi 1073741824 Apr 29 12:27 centos-8_yum_cache
-rw-r--r--. 1 nhosoi nhosoi 1073741824 Apr 29 12:27 centos-8_yum_varlib
```

richm:
> lgtm
> 
> Actually, I was going to ask a question related to this :). Why do we need to have the same looking files?
> 
> ```
> -rw-r--r--. 1 nhosoi nhosoi  169672704 Apr 29 12:27 centos-8.qcow2.snap

This is the writable snapshot image that uses the read-only `centos-8.qcow2` as the backing image.  We create a new one for each PR, so there is no reason to keep it (at least, the way we use it now).

> -rw-r--r--. 1 nhosoi nhosoi 1073741824 Apr 29 12:27 centos-8_yum_cache

This gets mounted at `/var/cache/dnf` in the image.  This is separate from the snapshot because it can be used without using the snapshot.  We could try to reuse this for multiple PRs, but for now it isn't worth it.

> -rw-r--r--. 1 nhosoi nhosoi 1073741824 Apr 29 12:27 centos-8_yum_varlib

This gets mounted at `/var/lib/dnf` in the image.  This is separate from the snapshot because it can be used without using the snapshot.  We could try to reuse this for multiple PRs, but for now it isn't worth it.

> ```

nhosoi:
Thanks a lot for the explanation!

PR #176, opened 2022-07-06
use collection parameter to runqemu

richm:


richm:
This is to use https://github.com/linux-system-roles/tox-lsr/pull/86

